PR #20, opened 2026-01-21
Encode ratchet sate for serialization

kaichaosun:
To persist the ratchet state, I got two options:
1. apps (upstream library) has the data storage, and libchat provides the binary data to resume the state.
2. managed storage in libchat, which will be in another PR.


jazzz:
The intention is always to be mindful of what data is being made available to whom. Previously allowing secret bytes to move out of its containing context would allow other modules to misuse this. Returning a slice instead is preferred.

 `as_bytes` is better - but I think this code could benefit from refactor in the future to improve the key handling.

Issue #18, opened 2026-01-15
Integrate Inbox with Conversations + API


jazzz:
## Problem 

Ensure that the API can access Inbox functions for creating bundles, creating conversations from bundles, and handling inbound messages. This Task is complete when a PrivateV1 convo can be initialized from the API using bundles generated from another client.

## Context
PR #13 provides functionality into Inbox, however it still needs to be integrated with the FFI api to be accessible

PR #24, opened 2026-01-23
Rust API - conversation creation

jazzz:
This PR adds the Rust side api for creating Private V1 conversations.

The api is implemented using Caller provided buffers, to avoid memory management across the FFI.

- Adds a convience function for the Caller to inform Context what size buffers it can expect.
  - If a ErrorCode::BufferExceeded would inform the caller to increase its buffer size and try again.
- Adds a `ConvoHandle` which acts as an opaque handle for conversations which are based over the FFI interface. Passing strings is costly and adds complexity.  
- Adds FFI functions for `create_intro_bundle`, `create_new_private_convo`

### Notes
- This PR is based upon #23, and merging this PR should not proceed until it is merged.
- This work is a is a precursor for https://github.com/logos-messaging/libchat/issues/18

### DIFF
--- a/conversations/src/context.rs
+++ b/conversations/src/context.rs
@@ -10,6 +10,9 @@ use crate::{
 
 pub use crate::inbox::Introduction;
 
+//Offset handles to make debuging easier
+const INITIAL_CONVO_HANDLE: u32 = 0xF5000001;
+
 /// Used to identify a conversation on the othersize of the FFI.
 type ConvoHandle = u32;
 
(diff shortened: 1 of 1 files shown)


conversations/src/context.rs
```diff
@@ -25,21 +32,32 @@ impl Context {
             _identity: identity,
             store: ConversationStore::new(),
             inbox,
+            buf_size: 0,
+            convo_handle_map: HashMap::new(),
+            next_convo_handle: 0xF5000001,
```
kaichaosun: Should be a constant defined outside of new, and maybe a little comment.

PR #30, opened 2026-01-28
Split storage crate to abstract database layer

kaichaosun:
What have changed:
- [x] remove feature gates and use sqlcipher by default
- [x] storage crate for database abstraction, this could be helpful if we want to introduce storage for other libchat modules or use a different database.

### DIFF
--- a/double-ratchets/examples/storage_demo.rs
+++ b/double-ratchets/examples/storage_demo.rs
@@ -2,17 +2,18 @@
 //!
 //! Run with: cargo run --example storage_demo -p double-ratchets
 
-use double_ratchets::{InstallationKeyPair, RatchetSession, RatchetStorage, hkdf::PrivateV1Domain};
+use double_ratchets::{InstallationKeyPair, RatchetSession, RatchetStorage};
+use tempfile::NamedTempFile;
 
 fn main() {
     println!("=== Double Ratchet Storage Demo ===\n");
 
-    ensure_tmp_directory();
-    let alice_db_path = "./tmp/double_ratchet_encrypted_alice.db";
-    let bob_db_path = "./tmp/double_ratchet_encrypted_bob.db";
+    let alice_db_file = NamedTempFile::new().unwrap();
+    let alice_db_path = alice_db_file.path().to_str().unwrap();
+    let bob_db_file = NamedTempFile::new().unwrap();
+    let bob_db_path = bob_db_file.path().to_str().unwrap();
+
     let encryption_key = "super-secret-key-123!";
-    let _ = std::fs::remove_file(alice_db_path);
-    let _ = std::fs::remove_file(bob_db_path);
 
     // Initial conversation with encryption
     {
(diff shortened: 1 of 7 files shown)


double-ratchets/src/storage/db.rs
```diff
@@ -0,0 +1,320 @@
+//! Ratchet-specific storage implementation.
+
+use std::collections::HashSet;
+
+use storage::{SqliteDb, StorageBackend, StorageError, params};
```
jazzz: I don't understand the pattern being used here. 

This PR moves DB storage to a different crate, but DR is still dependent upon Sqlite.

storage/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+/// Trait for storage backends.
+pub trait StorageBackend {
+    /// Initialize the storage (e.g., create tables).
+    fn init(&self) -> Result<(), StorageError>;
+
+    /// Execute a batch of SQL statements (for schema migrations).
+    fn execute_batch(&self, sql: &str) -> Result<(), StorageError>;
+}
```
jazzz: This Trait also doesn't appear to be used anywhere.

double-ratchets/src/storage/session.rs
```diff
@@ -1,58 +1,30 @@
...
 };
 
-use super::{SqliteStorage, StorageError};
+use super::RatchetStorage;
 
 /// A session wrapper that automatically persists ratchet state after operations.
 /// Provides rollback semantics - state is only saved if the operation succeeds.
 pub struct RatchetSession<'a, D: HkdfInfo + Clone> {
-    storage: &'a mut SqliteStorage,
+    storage: &'a mut RatchetStorage,
```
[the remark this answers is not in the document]
kaichaosun: I really like the idea, I also see some challenges on the transactional storage requirement and ratchet data highly coupled with storage record. Will continue exploring in this direction.

storage/src/lib.rs
kaichaosun: Removed

double-ratchets/examples/storage_demo.rs
```diff
@@ -1,85 +1,12 @@
...
 //!
-//! Run with: cargo run --example storage_demo --features storage
-//! For SQLCipher: cargo run --example storage_demo --features sqlcipher
+//! Run with: cargo run --example storage_demo -p double-ratchets
 
-#[cfg(feature = "storage")]
-use double_ratchets::{
-    InstallationKeyPair, RatchetSession, SqliteStorage, StorageConfig, hkdf::PrivateV1Domain,
-};
+use double_ratchets::{InstallationKeyPair, RatchetSession, RatchetStorage, hkdf::PrivateV1Domain};
```
jazzz: [Sand] Having to import hkdf::PrivateV1Domain seems cumbersome. It appears I missed this before. Implementing Default similar to `RachetState` would make this easier to use.

PR #32, opened 2026-01-29
Integrate DR into PrivateV1

jazzz:
This PR incorporates the double ratchet into the PrivateV1 by initializing sessions and wiring up Encrypt and Decrypt.

### DIFF
--- a/conversations/src/conversation/privatev1.rs
+++ b/conversations/src/conversation/privatev1.rs
@@ -10,6 +10,8 @@ use x25519_dalek::PublicKey;
 
 use crate::{
     conversation::{ChatError, ConversationId, Convo, Id},
+    errors::EncryptionError,
+    proto,
     types::AddressedEncryptedPayload,
     utils::timestamp_millis,
 };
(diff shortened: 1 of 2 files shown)


conversations/src/conversation/privatev1.rs
```diff
@@ -3,35 +3,92 @@ use chat_proto::logoschat::{
...
     }
 
-    fn encrypt(&self, frame: PrivateV1Frame) -> EncryptedPayload {
-        // TODO: Integrate DR
+    pub fn new_responder(seed_key: SecretKey, dh_self: InstallationKeyPair) -> Self {
+        Self {
+            // TODO: Danger - Fix double-ratchets types to Accept SecretKey
+            dr_state: RatchetState::init_receiver(seed_key.as_bytes().to_owned(), dh_self),
+        }
+    }
```
jazzz: `double-ratchets` requires a InstallationKey struct to be passed in.

As InstallationKey is just a thin wrapper around StaticKey @kaichaosun  would you be ok if I refactored double-ratchets to use a common keyType as the rest of the cryptography? It would be a StructWrapper around StaticKey which would have the effect:
- Keys can be passed without extra copies for clearer memory hygiene.
- Remove DR direct dependence on x25519 in our entire codebase. 
- Avoid Needless copies

We can add an AbstractType or Trait Generics if we really want it to be completely independent in the future.

conversations/src/conversation/privatev1.rs
```diff
@@ -3,35 +3,92 @@ use chat_proto::logoschat::{
...
+
+    fn encrypt(&mut self, frame: PrivateV1Frame) -> EncryptedPayload {
+        let encoded_bytes = frame.encode_to_vec();
```
kaichaosun: I think the name can be shorted like `to_bytes` or something similar

jazzz: what do you mean @kaichaosun ? I don't follow.

PR #34, opened 2026-01-30
feat: use just for quickly run tasks.

kaichaosun:


justfile
```diff
@@ -0,0 +1,13 @@
...
+
+test:
+  cargo test
+
+# Run all examples
+run-examples:
+  cargo run --example double_ratchet_basic
+  cargo run --example serialization_demo
+  cargo run --example storage_demo
+  cargo run --example out_of_order_demo
```
jazzz: Some of these examples are hidden behind feature flags. If we want people to be able to run examples easily then it would make sense to remove the `storage` feature as it doesn't seem that optional.

jazzz (commented):
I'm not against adding `Just` to make working with the repository easier to work with - especially as more complex tasks are added.

However this usecase seems unneeded. Examples exist to allow developers to see how to use the library and act as a starting point for exploration. Running all of them often seems counter productive.

Is there a reason I am missing?

justfile
kaichaosun: The `storage` feature has been removed in this PR: https://github.com/logos-messaging/libchat/pull/30, please review this PR too when you see fit.

justfile
```diff
@@ -0,0 +1,13 @@
+# The first receipt is the default task, run with `just`
+build:
+  cargo build
+
+test:
+  cargo test
+
+# Run all examples
+run-examples:
```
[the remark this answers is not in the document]
kaichaosun: Yes, CI will be added later. 
Tests is not clearly show the workflow like demos, that's why most rust libraries contains examples directory, which is good in may ways.
Running examples ensures examples runs and works, if you don't run it, how to know it still works?

PR #40, opened 2026-02-03
Refactor conversations

kaichaosun:
What changes:
- renaming traits related to ConvFactory
- move shared traits to common submodule
- inbox, dm(direct message), group split
- fix dead code
- isolate FFI into submodule
- restructure prost related bytes conversion


conversations/src/inbox/inbox.rs
```diff
@@ -142,6 +139,7 @@ impl Inbox {
         payload: proto::EncryptedPayload,
     ) -> Result<(SecretKey, proto::InboxV1Frame), ChatError> {
         let handshake = Self::extract_payload(payload)?;
+        let payload_bytes = handshake.payload.clone();
```
jazzz: Why clone here, instead of consuming the slice directly?

conversations/src/inbox/inbox.rs
```diff
@@ -184,12 +181,9 @@ impl Inbox {
         Ok(handshake)
     }
 
-    fn decrypt_frame(
-        enc_payload: proto::InboxHandshakeV1,
-    ) -> Result<proto::InboxV1Frame, ChatError> {
-        let frame_bytes = enc_payload.payload;
+    fn decrypt_frame(payload_bytes: prost::bytes::Bytes) -> Result<proto::InboxV1Frame, ChatError> {
```
jazzz: This change, weakens the type constraint on the function - Making auto complete more difficult and adding new failure modes to this function. 

Whats the motivation?

conversations/src/ffi/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod api;
```
jazzz: I'm not sure of the motivation here. Creating a module with a single file seems like added complexity.

conversations/src/proto.rs
```diff
@@ -7,3 +7,14 @@ pub use chat_proto::logoschat::invite::InvitePrivateV1;
...
+
+pub trait CopyBytes {
+    fn copy_to_bytes(&self) -> Bytes;
+}
+
+impl CopyBytes for PublicKey {
+    fn copy_to_bytes(&self) -> Bytes {
+        Bytes::copy_from_slice(self.as_bytes())
+    }
+}
```
jazzz: I think this is a good change. Importing all the types needed, and providing convience functions for encoding makes a lot of sense to me.

However noting for future expectations I'd like to see this trait removed in favor of NewType pattern implemented in #37.

This trait exists solely to get around the Orphan rule. There are better ways to do that.

jazzz (changes requested):
It's hard to understand the motivation behind these changes, and the reason why to change it now. 

I can see that you are focusing on better organization of the project, which is welcome as the project could use improvements - However I don't grok why this path is preferred. 

Clearly articulating the motivations would allow discussions to occur. 


Sticky points:
- Conversation != Session, they are not substitutable. 
- Trait Renaming doesn't accurately describe what is happening.
- Folder structure logic, I don't understand the vision.

conversations/src/dm/privatev1.rs
```diff
@@ -34,14 +35,14 @@ impl PrivateV1Convo {
     }
 }
 
-impl Id for PrivateV1Convo {
-    fn id(&self) -> ConversationId {
+impl HasConversationId for PrivateV1Convo {
+    fn id(&self) -> SessionId<'_> {
```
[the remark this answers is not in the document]
kaichaosun: Yeah, session seems not fit, I may find a new name, feel free to recomment new ones.
Conversation seems long, a short one would be good.

conversations/src/common.rs
```diff
@@ -0,0 +1,79 @@
...
+        encoded_payload: &[u8],
+    ) -> Result<(Box<dyn OutboundSession>, Vec<ContentData>), ChatError>;
+}
+
+pub trait OutboundSession: HasConversationId + Debug {
+    fn send_message(&mut self, content: &[u8])
+    -> Result<Vec<AddressedEncryptedPayload>, ChatError>;
+
+    fn remote_id(&self) -> String;
+}
```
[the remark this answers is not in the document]
kaichaosun: Is handle_frame specificly for handling incoming messages, and send_message for sending outgoing message?
Inbound ≈ Incoming, and Outbound ≈ Outgoing, I maybe totally wrong, not a native speaker :)

Session used here will be renamed to another word.

Conversation is confusing to me, because it focus on a large scenario, not specific capability can be easily reflected in code.

conversations/src/common.rs
```diff
@@ -0,0 +1,79 @@
...
+
+pub struct SessionRegistry {
+    sessions: HashMap<Arc<str>, Box<dyn OutboundSession>>,
+    handlers: HashMap<Arc<str>, Box<dyn InboundSessionHandler>>,
+}
+
+impl SessionRegistry {
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
kaichaosun: Sorry, I was focus on cleaning the code, haven't think about which is best place to integrate with persistence layer.

conversations/src/inbox/inbox.rs
jazzz: > so it don't get deleted
Then I would delete it.  I don't see much harm in that. Changing the signature to accept a more generic type (`Bytes`)loses context which is more harmful.

PR #42, opened 2026-02-05
Remove unneeded traits and warnings

jazzz:
This PR aims to clean up warnings in the workspace.

There currently are many warnings due to unfinished code-paths that result in an `unused` warnings.

Changes:
- Remove ConvoFactory trait - This is a premature optimization and not strictly needed. Removing for now.
- Remove Buffersize getters and setters. If needed for FFI it can be re-added.

The warnings which remain involve the inbound decryption path. These will be removed in the next PR which implements payload handling.


kaichaosun (approved):
💯

Issue #27, opened 2026-01-26
Introduction encoding


jazzz:
## Problem

Keys required to initiate an Inbox handshake need to be encoded need to be bundled for easy distribution. The bundle is distributed out-of-band so the introduction has UX implications.

Data required to be transferred.
- Installation PublicKey
- Ephemeral PublicKey
- Signature

osmaczko:
Binary Encoding Comparison

Criteria | TLV (Custom) | Protobuf | CBOR
-- | -- | -- | --
Efficient transport | 🟢 Excellent. Minimal overhead; tags and lengths only. Size-optimal if carefully designed. | Very good. Compact varint tags and lengths; small overhead relative to payload. | Very good. Compact binary representation; slightly higher overhead than TLV in some cases.
Schema evolution (forward & backward compatibility) | Manual. Fully possible but must be designed and enforced by the spec and implementations. | 🟢 Excellent. Built-in forward/backward compatibility via field numbers and unknown-field handling. | Good. Natural support via maps/arrays, but compatibility rules are convention-based rather than enforced.
Canonicality & signing semantics | Explicit. Canonicality and signing rules must be fully specified, but can be made unambiguous. | Non-canonical by default. Requires explicit rules (or exact-bytes signing) to avoid ambiguity. | 🟢 Strong. Deterministic CBOR provides built-in canonical encoding suitable for signing.
Implementation cost | High. Custom parsing, validation, and evolution rules in every language; higher risk of subtle bugs. | 🟢 Low. Code generation, stable APIs, and well-understood semantics across languages. | Medium. Libraries exist, but deterministic mode and conventions add complexity.
Tooling & ecosystem maturity | Low. No standard tooling; debugging and evolution are manual. | 🟢 Excellent. Mature, widely used tooling, codegen, linters, and long-term support. | Medium. Good library availability, but ecosystem is less uniform and less familiar to many developers.

jazzz:
Introductions are encoded and functioning

PR #43, opened 2026-02-05
feat: update introduction bundle encoding

osmaczko:
closes: https://github.com/logos-messaging/libchat/issues/27

### DIFF
--- a/conversations/src/inbox/introduction.rs
+++ b/conversations/src/inbox/introduction.rs
@@ -1,33 +1,92 @@
-use crypto::PrekeyBundle;
-use x25519_dalek::PublicKey;
+use base64::{Engine, engine::general_purpose::URL_SAFE_NO_PAD};
+use chat_proto::logoschat::intro::IntroBundle;
+use crypto::Ed25519Signature;
+use prost::Message;
+use rand_core::{CryptoRng, RngCore};
+use x25519_dalek::{PublicKey, StaticSecret};
 
 use crate::errors::ChatError;
 
-/// Supplies remote participants with the required keys to use Inbox protocol
-pub struct Introduction {
-    pub installation_key: PublicKey,
-    pub ephemeral_key: PublicKey,
+const BUNDLE_PREFIX: &str = "logos_chatintro_1_";
+
+fn intro_binding_message(ephemeral: &PublicKey) -> Vec<u8> {
+    let mut message = Vec::with_capacity(BUNDLE_PREFIX.len() + 32);
+    message.extend_from_slice(BUNDLE_PREFIX.as_bytes());
+    message.extend_from_slice(ephemeral.as_bytes());
+    message
 }
 
-impl From<PrekeyBundle> for Introduction {
-    fn from(value: PrekeyBundle) -> Self {
-        Introduction {
-            installation_key: value.identity_key,
-            ephemeral_key: value.signed_prekey,
+pub(crate) fn sign_intro_binding<R: RngCore + CryptoRng>(
+    secret: &StaticSecret,
+    ephemeral: &PublicKey,
+    rng: R,
+) -> Ed25519Signature {
+    let message = intro_binding_message(ephemeral);
+    crypto::xeddsa_sign(secret, &message, rng)
+}
+
+pub(crate) fn verify_intro_binding(
+    pubkey: &PublicKey,
+    ephemeral: &PublicKey,
+    signature: &Ed25519Signature,
+) -> Result<(), crypto::SignatureError> {
+    let message = intro_binding_message(ephemeral);
+    crypto::xeddsa_verify(pubkey, &message, signature)
+}
+
+/// Supplies remote participants with the required keys to use Inbox protocol
+pub struct Introduction {
+    installation_key: PublicKey,
+    ephemeral_key: PublicKey,
+    signature: Ed25519Signature,
+}
+
+impl Introduction {
+    /// Create a new `Introduction` by signing the ephemeral key with the installation secret.
+    pub(crate) fn new<R: RngCore + CryptoRng>(
+        installation_secret: &StaticSecret,
+        ephemeral_key: PublicKey,
+        rng: R,
+    ) -> Self {
+        let installation_key = installation_secret.into();
+        let signature = sign_intro_binding(installation_secret, &ephemeral_key, rng);
+        Self {
+            installation_key,
+            ephemeral_key,
+            signature,
         }
+    }
+
+    pub fn installation_key(&self) -> &PublicKey {
+        &self.installation_key
+    }
+
+    pub fn ephemeral_key(&self) -> &PublicKey {
+        &self.ephemeral_key
+    }
+
+    pub fn signature(&self) -> &Ed25519Signature {
+        &self.signature
     }
 }
 
 impl From<Introduction> for Vec<u8> {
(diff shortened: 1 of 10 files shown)


conversations/src/inbox/introduction_encoding.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Signature
+
+The `signature` field proves that the holder of the installation private key
+authorized the ephemeral public key for use in key agreement.
+
+**Algorithm:** [XEdDSA][xeddsa] — Ed25519-compatible signatures produced from
+X25519 key pairs. This allows the same X25519 installation key to be used for
+both Diffie-Hellman key agreement and signing, without maintaining a separate
+Ed25519 identity key.
```
jazzz: I don't think this belongs in the Spec for encoding. This is chat protocol specific, best to not repeat information in multiple places. Recommend removing

conversations/src/inbox/introduction_encoding.md
```diff
@@ -0,0 +1,204 @@
...
+signature           : <64-byte XEdDSA signature over
+                       "logos_chatintro_v1_bind" || ephemeral_pubkey>
+```
+
+Encoding steps:
+
+1. Construct the `IntroBundle` protobuf message with the three fields above.
+2. Serialize using proto3 encoding (134 bytes).
+3. Encode as base64url without padding (179 characters).
+4. Prepend the version prefix:
```
jazzz: [Sand] The ought to be listed clearly enough. If a reader cannot easily follow along, then the instructions ought to be made clearer.

conversations/src/inbox/introduction_encoding.md
```diff
@@ -0,0 +1,204 @@
...
+
+The underscore was chosen because it is a printable ASCII character that is
+safe in URLs, filenames, and plain text without requiring escaping. Delimiters
+outside this category (e.g., `.` or `:`) may require percent-encoding in
+certain URL contexts, undermining the transport-safety goal.
+
+## V1
+
+A V1 Introduction Bundle carries the cryptographic material needed for an
+[X3DH][x3dh] key agreement:
```
jazzz: [Sand] This seems hyper specific. The spec should theoretically work for another protocol stack that matches the same structure.

Since this protocol treats the internal bytes as blobs then they theoretically could be anything. 

High Level I see this spec being referred to by a spec like Inbox. Inbox defines the keys and types, it then references this spec for its encoding.

conversations/src/inbox/introduction_encoding.md
```diff
@@ -0,0 +1,204 @@
...
+| `payload`   | Binary-to-text encoded payload (see below).                       |
+
+Fields are separated by `_` (underscore).
+
+### Delimiter Choice
+
+The `_` character is present in the alphabets of common binary-to-text
+encodings (hex, base32, base64url) and may therefore appear inside
+`<payload>`. Parsers MUST split the encoded string on the **first three** `_`
+characters; everything after the third `_` is the payload verbatim.
```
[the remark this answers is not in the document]
osmaczko: What you meant is something like `logos_chatintro_1~<payload>`, where `_` separates preamble fields and `~` separates preamble from payload. Then parsing becomes: split on `~` once -> left side is preamble, right side is payload. No "first three" counting. Did I get that right?                                                                 
                                                                                                                                                                       
That's a good suggestion. However, imo, it adds complexity for marginal benefit. The "split on first 3 underscores" rule is simple and unambiguous, typically one line of code in any language. Two delimiters means two characters to document, two things to validate, and slightly more visual noise.

I'd lean toward keeping the single delimiter if you're fine with it.

conversations/src/inbox/introduction_encoding.md
osmaczko: I agree, removed.

conversations/src/inbox/introduction.rs
```diff
@@ -1,64 +1,180 @@
...
-use x25519_dalek::PublicKey;
+use prost::Message;
+use rand_core::{CryptoRng, RngCore};
+use x25519_dalek::{PublicKey, StaticSecret};
 
 use crate::errors::ChatError;
 
+const BUNDLE_PREFIX: &str = "logos_chatintro_1_";
+
+const INTRO_DOMAIN_SEPARATOR: &[u8] = b"logos_chatintro_v1_bind";
```
jazzz: So close but so different. Any reason to not use the same prefix?

crypto/src/xeddsa_sign.rs
```diff
@@ -0,0 +1,118 @@
...
+
+/// Error type for signature verification failures.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SignatureError;
+
+impl std::fmt::Display for SignatureError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "signature verification failed")
+    }
+}
```
jazzz: [Sand] take a look at the `thiserror` library for error handling.

conversations/src/inbox/introduction.rs
```diff
@@ -1,64 +1,180 @@
+use base64ct::{Base64UrlUnpadded, Encoding};
```
jazzz: This line set off alarm bells. Is key material being encoded in base64?

Is a constant time operation required here or being cautious?

conversations/src/inbox/introduction.rs
```diff
@@ -1,64 +1,180 @@
...
+    message.extend_from_slice(identity.as_bytes());
+    message.extend_from_slice(ephemeral.as_bytes());
+    message
+}
+
+pub(crate) fn sign_intro_binding<R: RngCore + CryptoRng>(
+    secret: &StaticSecret,
+    ephemeral: &PublicKey,
+    rng: R,
+) -> [u8; 64] {
```
jazzz: [Pebble] I'd create a Newtype using `struct Ed25519Signature { sig: [u8; 64] }`

That way the context is bound with the datatype as well as allowing convenience functions to be added in the future.

conversations/src/inbox/introduction.rs
```diff
@@ -1,64 +1,180 @@
...
 use crate::errors::ChatError;
 
+const BUNDLE_PREFIX: &str = "logos_chatintro_1_";
+
+const INTRO_DOMAIN_SEPARATOR: &[u8] = b"logos_chatintro_v1_bind";
+
+fn intro_binding_message(identity: &PublicKey, ephemeral: &PublicKey) -> Vec<u8> {
+    let mut message = Vec::with_capacity(INTRO_DOMAIN_SEPARATOR.len() + 64);
+    message.extend_from_slice(INTRO_DOMAIN_SEPARATOR);
+    message.extend_from_slice(identity.as_bytes());
```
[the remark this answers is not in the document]
osmaczko: True, removed identity key from the message.

conversations/src/inbox/introduction.rs
osmaczko: The encoded data has no secret material, using regular base64 would be fine as well as there's nothing to leak via timing. I used constant-time as a reasonable default, as far as I know, there is no meaningful penalty.

osmaczko:
@jazzz thanks for review! I've addressed your comments, please re-check :pray:

PR #52, opened 2026-02-10
fix: make ci checks pass

osmaczko:


jazzz (approved):
These are changes that have been bugging me. Thanks for taking care of them.

Pre-approving, but I'd like to see a plan for exposing conversation_ids() over FFI to make sure that work doesn't get lost.

PR #61, opened 2026-02-23
chore(nim-bindings): replace dynlib dlopen with plain importc

osmaczko:
The dynlib pragma hard-coded a library path and resolved it via dlopen() at runtime, preventing static linking and forcing a specific load-time path. Using bare {.importc.} lets consumers choose: link liblibchat dynamically at link time (--passL:-llibchat) or link it statically into their binary.

PR #62, opened 2026-02-24
Rust -> Nim ABI 

jazzz:
# Problem
From what I can see, there is an ABI incompatibility between NIM and safer_ffi. The issue appears to be caused by a difference in the calling conventions used by nim. 

From initial testing. safer_ffi and c-binding interfaces use a convention called `sret` to handle the return of large structs. Where nim overwrites the signature to include an extra `out` parameter and does not adhere to the `sret` convention. AARM64 defines this behavior in[ `aapcs64`](https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#result-return)

This was surfaced by https://github.com/logos-messaging/libchat/pull/61 which exposed a SegFault. Having spent more time that I'd like digging around in compiler code, I was unable to find concrete reason of why this was able to function before.

The reason for the segfault on my Apple Silicone mac is clear. According to aapcs64, the rust function expected the caller to allocate memory and write the address to register `x8`. As nim did not follow this convention, it never initialized register `x8`, resulting to undefined behavior when the memory location is accessed. 

# Solution

This PR removes the calling convention ambiguity and forces the rust functions to follow nim [convention ](https://forum.nim-lang.org/t/2665). 

Other approaches considered:
- Using a shim layer to convert between nim and `sret` conventions. This produces yet another layer and seemed more problematic.

# Notes
- This approach does mean that calling via the FFI API to languages like Python over C_API won't work, as the files allow the non-standard nim convention
- Also improves the build script for the bindings example.

### DIFF
--- a/conversations/src/lib.rs
+++ b/conversations/src/lib.rs
@@ -25,7 +25,11 @@ mod tests {
         let mut raya = create_context("raya".into());
 
         // Raya Creates Bundle and Sends to Saro
-        let intro_result = create_intro_bundle(&mut raya);
+        let mut intro_result = CreateIntroResult {
+            error_code: -99,
+            intro_bytes: safer_ffi::Vec::EMPTY,
+        };
+        create_intro_bundle(&mut raya, &mut intro_result);
         assert!(is_ok(intro_result.error_code));
 
         let raya_bundle = intro_result.intro_bytes.as_ref();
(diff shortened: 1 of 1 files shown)


osmaczko:
Thanks :pray:!

I’ve also tried the alternative approach proposed: https://github.com/logos-messaging/libchat/pull/64. The shim layer is not that complex and allows the Rust API to remain idiomatic

Another reason for preferring this approach is that `nim-bindings` should not be the responsibility of the Rust library itself. They belong to a separate layer, i.e. should either live in a dedicated repository or be maintained on the consumer side, for example in `logos-chat`. If that’s the case, then adapting the Rust FFI to accommodate `nim-bindings` feels wrong.

Let me know what are your thoughts.

osmaczko (approved):

PR #63, opened 2026-02-24
fix: sigsegv on starting a conversation

plopezlpz:
Getting this errors on `logos-chatsdk-ui` when trying to start a conversation:

"SIGSEGV: Illegal storage access. (Attempt to read from nil?)"


jazzz (commented):
This issue currently has two proposed solutions: #62 and #64

I don't immediately see how `{.bycopy.}` directly solves the issue. Perhaps I am missing something though.

PR #64, opened 2026-02-24
fix(nim-bindings): bridge Nim/C ABI mismatch via C shim layer

osmaczko:
Nim's code generator transforms function signatures involving large structs in two ways that conflict with the standard C ABI:

  - Return values of large structs (> register size): Nim emits a void function with an explicit out-pointer appended as the *last* argument. The standard x86-64 SysV ABI passes the hidden return pointer in RDI (before the real arguments); ARM64 aapcs64 uses X8. Calling Rust directly from Nim therefore puts the pointer in the wrong register / stack slot on both architectures, causing crashes.

  - Large struct parameters (> ~24 bytes): Nim passes a pointer rather than copying bytes on the stack / into registers as the C ABI expects.

This commit introduces a thin C shim (nim_shims.c) that acts as a translation layer:

  - Each nim_* wrapper is declared with a Nim-compatible signature, so Nim calls it correctly by its own rules.
  - Inside the wrapper the C compiler calls the real Rust-exported function using the standard C ABI, inserting the correct hidden- pointer placement and stack-copy behaviour for the current platform.

As a result:
  - The Rust API stays standard C ABI (return T by value; destroy takes *mut T, which is pointer-sized and matches Nim's large-param transform).
  - Other language bindings (C, Swift, Go, …) call Rust directly without any shim — the standard ABI is preserved for them.
  - The fix is correct on both x86-64 and ARM64 without any architecture-specific code in Nim or Rust.

Changes:
  - nim-bindings/src/nim_shims.c: C bridge with nim_* wrappers for all create/handle/installation_name and destroy functions
  - nim-bindings/src/bindings.nim: {.compile: "nim_shims.c"}, proc signatures use natural return-by-value form, importc names point to the nim_* shims
  - nim-bindings/src/libchat.nim: call sites use natural let binding form; destroy calls pass addr res (ptr T)
  - conversations/src/api.rs: destroy functions take *mut T so Nim's large-param-to-pointer transform is satisfied without a stack copy

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,3 +36,18 @@ jobs:
       - run: rustup update stable && rustup default stable
       - run: rustup component add rustfmt
       - run: cargo fmt --all -- --check
+
+  nim-bindings-test:
+    name: Nim Bindings Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: rustup update stable && rustup default stable
+      - name: Install Nim via choosenim
+        run: |
+          curl https://nim-lang.org/choosenim/init.sh -sSf | sh -s -- -y
+          echo "$HOME/.nimble/bin" >> $GITHUB_PATH
+      - run: nimble install -dy
+        working-directory: nim-bindings
+      - run: nimble pingpong
+        working-directory: nim-bindings
--- a/conversations/src/api.rs
+++ b/conversations/src/api.rs
@@ -196,8 +196,10 @@ pub struct CreateIntroResult {
 
 /// Free the result from create_intro_bundle
 #[ffi_export]
-pub fn destroy_intro_result(result: CreateIntroResult) {
-    drop(result);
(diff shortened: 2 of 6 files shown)


jazzz (commented):
Oof. I don't like this or #62.

Originally caller provided buffer approach was abandoned, in favor of safer_ffi's better flexibility and type safety. 

This issue with nim abi is diminishing the benefit of safer_ffi.

The Split-Ownership model, where the Caller owns the outer struct and rust owns the inner members is going to cause issues. Returning results by value and then freeing by reference lacks symmetry which makes it harder to reason about. 

I don't have any strict blockers for this approach (assuming comments are addressed) however I'm starting to wonder if a Box allocated model would be cleaner. Heap-Allocate every call via `-> Box<T>`. Memory is completely managed by rust, and since the return type is a boxed pointer all issues with the ABI are sidestepped.

osmaczko:
Arguments given in https://github.com/logos-messaging/libchat/pull/62#issuecomment-3953599751 convinced me. We can get back to this solution once it is needed (i.e. more bindings).

PR #67, opened 2026-02-27
Persist identity

kaichaosun:
What changes:
- [x] integrate the storage crate for sqlite access
- [x] migrations and run helpers
- [x] save and load identity
- [x] expose ffi with app provided storage path


plopezlpz (approved):

conversations/src/storage/db.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// Applies all migrations and returns the storage instance.
+    fn run_migrations(db: SqliteDb) -> Result<Self, StorageError> {
+        migrations::apply_migrations(db.connection())?;
+        Ok(Self { db })
+    }
+
+    // ==================== Identity Operations ====================
+
+    /// Saves the identity (secret key).
+    pub fn save_identity(&mut self, identity: &Identity) -> Result<(), StorageError> {
```
jazzz: [Sand] Clearing up the language used in this library: `Installation` represent keys for this given instance, where Identity represents a group of installations tied to a set of Installations.

conversations/src/storage/db.rs
```diff
@@ -0,0 +1,96 @@
...
+        self.db.connection().execute(
+            "INSERT OR REPLACE INTO identity (id, name, secret_key) VALUES (1, ?1, ?2)",
+            params![
+                identity.get_name(),
+                identity.secret().DANGER_to_bytes().as_slice()
```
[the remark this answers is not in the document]
jazzz: There is also a deprecation notice on `DANGER_to_bytes` which we may need to re-discuss. 

The intention is that the conversations crate will eventually not have any access to key material, and all operations on secrets bytes would be localized to a single location such as crate crypto. This makes external reviews and safety checks straight forward by establishing an invariant that crates cannot have unsafe access to key material.

Storage poses a problem with this approach and will need to be need to be considered more fully.

conversations/src/storage/mod.rs
```diff
@@ -0,0 +1,7 @@
+//! Storage module for persisting chat state.
+
+mod db;
+mod migrations;
+pub(crate) mod types;
+
+pub(crate) use db::ChatStorage;
```
jazzz: This project (with the exception of double-ratchets) follows the Rust 2018 guidance -  https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html

conversations/src/ffi/utils.rs
```diff
@@ -0,0 +1,13 @@
+use safer_ffi::prelude::*;
+
+#[derive_ReprC]
+#[repr(C)]
+pub struct CResult<T: ReprC, Err: ReprC> {
+    pub ok: Option<T>,
+    pub err: Option<Err>,
+}
```
jazzz: From a consistency POV, api.rs uses the safer_ffi to create unique types per each api_call. I'm not saying that is the best approach, however this does represent a different approach of using generic containers as return values.

conversations/src/storage/types.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub name: String,
+    /// The secret key bytes (32 bytes).
+    pub secret_key: [u8; 32],
+}
+
+impl From<IdentityRecord> for Identity {
+    fn from(record: IdentityRecord) -> Self {
+        let name = record.name.clone();
+        let secret = PrivateKey::from(record.secret_key);
+        Identity::from_secret(name, secret)
```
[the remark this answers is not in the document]
jazzz: 💯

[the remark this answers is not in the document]
kaichaosun: Move name did not pass the compiler,
```
cannot move out of type `IdentityRecord`, which implements the `Drop` trait
```

conversations/src/storage/mod.rs
kaichaosun: I don't see why we should follow this "guidance".
I have randomly searched a few top repos in github, all of my searched repos includes `mod.rs`.
https://github.com/EvanLi/Github-Ranking/blob/master/Top100/Rust.md

conversations/src/ffi/utils.rs
kaichaosun: Exactly, that's why I put in ffi/utils, I hope it can be reused in other apis, so that we can have uniform apis and reused code.

jazzz (approved):
With the amount of changes occurring I think this PR is one of the victims. 

Theres good work in here, in particular i see:
- a commitment to secret hygiene, by ensuring memory is zeroized during storage.
- Focus on data lifecycle and migration management.

I won't block merging this as is, if you'd like to close it out.

PR #69, opened 2026-03-22
Group integration example

jazzz:
This is an Example of what is required to get MLS groupchats integrated into Conversations API.


jazzz:
Closed in favor of #92

PR #70, opened 2026-03-22
Flatten Repos

jazzz:
This PR prepares for the combination of 'libchat' and 'logos-chat'.

- Use 'crates' folder approach to separate rust crates from other code.
- Use separate `crate` folders to maintain separation between protocol core and platform implementation details.
- Expose libchat::Context so it can be used directly without using ffi.

Work deferred:
- Removing FFI and bindings.


osmaczko (approved):
LGTM.

For the record, quoting an offline discussion below.

I proposed adding an `internal` subdirectory under `core_crates` due to concerns that it could otherwise be too easy to take shortcuts, for example by directly using `double-ratchets` logic or other low-level implementation details in logos-chat. This would risk creating tight coupling through reliance on libchat internals rather than its intended public interfaces.

Jazz pointed out that introducing an `internal` directory in libchat would be undesirable in the long term because:

1) It is non-standard.
2) It contradicts the concept of isolated, reusable compilation units.

As a result, such a structure would likely need to be removed in the future when the libchat repository is extracted, which would in turn break existing paths and introduce unnecessary churn.

Cargo.toml
```diff
@@ -3,15 +3,16 @@
...
 
 members = [
-    "conversations",
-    "crypto",
-    "double-ratchets",
-    "storage",
+    "core_crates/conversations",
+    "core_crates/crypto",
+    "core_crates/double-ratchets",
+    "core_crates/storage",
```
kaichaosun: [Discussion] Not sure if storage should live here, it seems platform specific, also very coupled with conversations and other core crates, we can make it live here now, and decide later.

kaichaosun (commented):
The change looks good.
But I'm not sure if the new 2 layer directory abstraction brings more clarity, or over designed in current stage. For example, if chat-crates been moved out in the future, do we need to remove "core-crates" folder too? 

IMO a good crate name and readme within one crate should be good enough to help with review and dev usage.

Cargo.toml
jazzz: Absolutely Agreed. I think you already have a proposal for this. However I am trying to avoid any friction in this PR and only updated the layout as it stands today. 

Having the two crates folders helps to be clear about the intention of how the repo will be split in a few months when its needed to isolate the core crates from the platform specific implementations. Hopefully the PR that splits will be straight forward given they are already split and agreed upon.

plopezlpz (approved):

kaichaosun (approved):

Issue #68, opened 2026-03-18
Make modules system consistent


kaichaosun:
There are 2 styles when introduce modules in Rust, following are the structure with a chat module:

Old style (with mod.rs), 
```
src
 |- chat
        |- mod.rs
        |- other.rs
 |- lib.rs
 |- xyz.rs
```

New style:

```
src
 |- chat
        |- other.rs
 |- lib.rs
 |- xyz.rs
 |- chat.rs
```

Signal your preference so we can decide which styles to use. After we have all the dev signals, change the codebase to use the agreed style.

PR #72, opened 2026-03-25
refactor: use new styling for conversations crate without mod.rs

kaichaosun:
This is the follow up PR to clean the repo to use new module styling.

More context: https://github.com/logos-messaging/libchat/issues/68


jazzz (approved):
👍

Issue #71, opened 2026-03-23
Chat Interface Client


jazzz:
With this repo becoming directly consumable, a client API needs to be implemented. This work is complete when there is a method for applications to create functioning chat applications directly in Rust. 

- Expose functionality for initializing a client with the appropriate DeliveryService, AuthenticationService, StorageService
- Expose client management functions for creating conversations
- Expose Conversation functionality for sending, receiving messages

### Context

- Previously this functionality was provided by logos-chat::Client.

jazzz:
ChatClient is working in the bin/chat-cli

PR #73, opened 2026-03-26
feat: implement rust client

osmaczko:
Implement a `client` crate that wraps the `libchat` context behind a
simple `ChatClient<D>` API. The delivery strategy is pluggable via a
`DeliveryService` trait, with two implementations provided:

- `InProcessDelivery` — shared `MessageBus` for single-process tests
- `CDelivery` — C function-pointer callback for the FFI layer

Add a `client-ffi` crate that exposes the client as a C API via
`safer-ffi`. A `generate-headers` binary produces the companion C
header.

Include two runnable examples:
- `examples/in-process` — Alice/Bob exchange using in-process delivery
- `examples/c-ffi` — same exchange written entirely in C; smoketested
under valgrind (to catch memory leaks) in CI

iterates: https://github.com/logos-messaging/libchat/issues/71

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -1,22 +1,93 @@
-use chat_sqlite::StorageConfig;
-use libchat::ChatError;
-use libchat::ChatStorage;
-use libchat::Context;
+use libchat::{
+    AddressedEnvelope, ChatError, ChatStorage, ContentData, Context, ConversationIdOwned,
+    Introduction, StorageConfig,
+};
 
-pub struct ChatClient {
+use crate::{delivery::DeliveryService, errors::ClientError};
+
+pub struct ChatClient<D: DeliveryService> {
     ctx: Context<ChatStorage>,
+    delivery: D,
 }
 
-impl ChatClient {
-    pub fn new(name: impl Into<String>) -> Self {
-        let store =
-            ChatStorage::new(StorageConfig::InMemory).expect("in-memory storage should not fail");
+impl<D: DeliveryService> ChatClient<D> {
+    /// Create an in-memory, ephemeral client. Identity is lost on drop.
+    pub fn new(name: impl Into<String>, delivery: D) -> Self {
+        let store = ChatStorage::in_memory();
         Self {
             ctx: Context::new_with_name(name, store),
+            delivery,
         }
     }
 
-    pub fn create_bundle(&mut self) -> Result<Vec<u8>, ChatError> {
-        self.ctx.create_intro_bundle()
+    /// Open or create a persistent client backed by `StorageConfig`.
+    ///
+    /// If an identity already exists in storage it is loaded; otherwise a new
+    /// one is created and saved.
+    pub fn open(
+        name: impl Into<String>,
+        config: StorageConfig,
+        delivery: D,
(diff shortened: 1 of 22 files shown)


crates/client/src/client.rs
```diff
@@ -1,18 +1,90 @@
...
+    ) -> Result<Option<ContentData>, ClientError<D::Error>> {
+        self.ctx.handle_payload(payload).map_err(Into::into)
+    }
+
+    fn dispatch_all(
+        &mut self,
+        envelopes: Vec<AddressedEnvelope>,
+    ) -> Result<(), ClientError<D::Error>> {
+        for env in envelopes {
+            self.delivery.deliver(env).map_err(ClientError::Delivery)?;
```
kaichaosun: Wondering how is the delivery error propagated to core/conversations so that the state is consistent and recovered from error happens?

crates/client/src/delivery_in_process.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl MessageBus {
+    /// Returns a cursor positioned at the beginning of `address`.
+    /// The cursor will see all messages — past and future.
+    pub fn subscribe(&self, address: &str) -> Cursor {
+        Cursor {
+            bus: self.clone(),
+            address: address.to_string(),
+            pos: 0,
```
kaichaosun: If starts from 0, probably not subscribe, or semantics like `get_history`.

crates/client/src/delivery.rs
```diff
@@ -0,0 +1,6 @@
+use libchat::AddressedEnvelope;
+
+pub trait DeliveryService {
+    type Error: std::fmt::Debug;
+    fn deliver(&mut self, envelope: AddressedEnvelope) -> Result<(), Self::Error>;
+}
```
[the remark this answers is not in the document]
osmaczko: Good point, will keep that in mind.

crates/client/src/delivery.rs
```diff
@@ -0,0 +1,6 @@
+use libchat::AddressedEnvelope;
+
+pub trait DeliveryService {
+    type Error: std::fmt::Debug;
+    fn deliver(&mut self, envelope: AddressedEnvelope) -> Result<(), Self::Error>;
```
jazzz: [Dust] I would probably call this `publish`.

jazzz (commented):
I think this PR is a good step forward. 

I don't see any blockers to getting this merged. The code is well organized and documented. I think we'll need to move some of the functionality around in order to improve DX and maintainability, but those don't need to be done now. 

Great Work.

jazzz (approved):

kaichaosun (approved):

crates/client/src/delivery_in_process.rs
```diff
@@ -0,0 +1,112 @@
...
+impl InProcessDelivery {
+    /// Create a delivery service backed by `bus`.
+    pub fn new(bus: MessageBus) -> Self {
+        Self(bus)
+    }
+
+    /// Returns a reference to the underlying [`MessageBus`].
+    pub fn bus(&self) -> &MessageBus {
+        &self.0
+    }
```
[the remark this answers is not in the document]
osmaczko: Good point, thanks

PR #75, opened 2026-03-27
feat: prepare for logos module consumption

plopezlpz:
- add flake.nix for Nix builds (cdylib + header generation)
- switch crate-type from dylib to cdylib for C FFI
- replace placeholder delivery addresses with real derived addresses
- switch FFI inputs from owned to borrowed
- expose local_delivery_address() so caller knows where to subscribe


jazzz (commented):
Currently I have many questions. I think the PR description could use more details on "WHY" these changes are occurring. 

Contextually. I think we'd like to see minimal to zero changes in the core for supporting modules. Specifically the integration point would be https://github.com/logos-messaging/libchat/issues/71

PR #81, opened 2026-04-06
Use inbox to handle save and load ephemeral key from store

kaichaosun:
For save and load ephemeral key, it's safe to move to inbox.
For delete ephemeral key, it depends on if successfully dispatch to convo, so delete op still lives in context.


jazzz (approved):
I think Interior mutability pattern makes sense here.

No blockers from my perspective, this looks good

PR #82, opened 2026-04-08
Move double ratchet storage operations to PrivateV1

kaichaosun:
**Why this changes?**

Based on the previous discussion, we want to put the store operations inside each module out of context, as each module has more knowledge on how to deal with storage.

**What changes:**
- add store dependency in `PrivateV1`
- load private v1 convo from store
- persist private v1 vonvo
- save ratchet state in handle message and handle frame
- clear context

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,8 +29,8 @@ target
 
 # Temporary data folder
 tmp
-
 .DS_Store
+justfile
 
 # Generated C headers (produced by `make` in examples/c-ffi; do not commit)
 crates/client-ffi/client_ffi.h
--- a/.gitignore
+++ b/.gitignore
@@ -29,8 +29,8 @@ target
 
 # Temporary data folder
 tmp
+
 .DS_Store
-justfile
 
 # Generated C headers (produced by `make` in examples/c-ffi; do not commit)
 crates/client-ffi/client_ffi.h
--- a/core/conversations/src/context.rs
+++ b/core/conversations/src/context.rs
@@ -2,7 +2,6 @@ use std::sync::Arc;
 use std::{cell::RefCell, rc::Rc};
 
 use crypto::Identity;
-use double_ratchets::{RatchetState, restore_ratchet_state};
 use storage::{ChatStore, ConversationKind};
 
 use crate::{
@@ -181,7 +180,7 @@ impl<T: ChatStore> Context<T> {
         Ok(intro.into())
     }
 
-    /// Loads a conversation from DB by constructing it from metadata + ratchet state.
+    /// Loads a conversation from DB by constructing it from metadata.
     fn load_convo(&self, convo_id: ConversationId) -> Result<Conversation<T>, ChatError> {
         let record = self
             .store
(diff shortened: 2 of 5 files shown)


core/conversations/src/conversation/privatev1.rs
```diff
@@ -55,23 +56,37 @@ impl BaseConvoId {
...
-pub struct PrivateV1Convo {
+pub struct PrivateV1Convo<S: ConversationStore + RatchetStore> {
     local_convo_id: String,
     remote_convo_id: String,
     dr_state: RatchetState,
+    store: Rc<RefCell<S>>,
 }
 
-impl PrivateV1Convo {
+impl<S: ConversationStore + RatchetStore> PrivateV1Convo<S> {
```
jazzz, line 66: [Dust] Creating an empty supertrait bound, would make the code easier to follow along, and easier to maintain.

core/conversations/src/conversation/privatev1.rs
```diff
@@ -86,10 +101,15 @@ impl PrivateV1Convo {
...
+            store,
         }
     }
 
-    pub fn new_responder(seed_key: SymmetricKey32, dh_self: &PrivateKey) -> Self {
+    pub fn new_responder(
+        seed_key: SymmetricKey32,
+        dh_self: &PrivateKey,
+        store: Rc<RefCell<S>>,
+    ) -> Self {
```
jazzz, line 112: [Sand] store should come before the other parameters. 

Parameters should be ordered most-stable to least-stable. Such as:
**self** — the actor
**context/deps** — what the actor needs to function (an allocator, a fmt handle) — rarely changes between calls
**the main data** — the thing being transformed or inspected
**options/flags** — tweaks to the operation
**output buffer** — where to write results, if not returned directly

core/conversations/src/context.rs
```diff
@@ -94,7 +93,7 @@ impl<T: ChatStore> Context<T> {
             .map(|p| p.into_envelope(remote_id.clone()))
             .collect();
 
-        let convo_id = self.persist_convo(&convo)?;
+        let convo_id = convo.persist()?;
```
jazzz, line 96: If conversations know how to persist themselves, whats the advantage to calling persist externally from the context? Seems like its separating storage responsibilities across two different layers.

core/conversations/src/context.rs
```diff
@@ -181,8 +180,8 @@ impl<T: ChatStore> Context<T> {
         Ok(intro.into())
     }
 
-    /// Loads a conversation from DB by constructing it from metadata + ratchet state.
-    fn load_convo(&self, convo_id: ConversationId) -> Result<Conversation, ChatError> {
+    /// Loads a conversation from DB by constructing it from metadata.
+    fn load_convo(&self, convo_id: ConversationId) -> Result<Conversation<T>, ChatError> {
```
jazzz: [Dust] `T` implies that this Conversation has is an operand type.  Better off choosing, `S` for Store or `Store` for Store, as there will be more Generic parameters.

.gitignore
```diff
@@ -31,3 +31,5 @@ target
 tmp
 
 .DS_Store
+
+justfile
```
jazzz: [Pebble] Can you exclude this instead. This project may have justfiles in the future which would cause a conflict.

core/conversations/src/context.rs
jazzz: Specifically, this would call persist regardless of whether the state needs to be updated. Removing persist from Conversation Trait would mean that clients didn't manage internal state at all.

jazzz (approved):
Overall Love the direction here. Conversations becoming more autonomous. 

Few comments around further clean up, and simplification. 

Approved assuming the Justfile Pebble is cleaned up.

.gitignore
kaichaosun: If the project include justfile in the future, remove it then.
I have been using justfile for this project heavily, not ignore it makes the git commit flow really bad.

core/conversations/src/context.rs
kaichaosun: Looks like a good idea, will refactor it later to keep this RP small.
There is a lot of discussion spaces around here, like clean up state when error happens, and how to resume for error scenarios, etc. Such requirements will be much clear when MVPs start releasing.

.gitignore
jazzz: > not ignore it makes the git commit flow really bad.

I asked if you could use [git exlcude ](https://docs.github.com/en/get-started/git-basics/ignoring-files#excluding-local-files-without-creating-a-gitignore-file) instead.

core/conversations/src/conversation/privatev1.rs
kaichaosun: I have think about introduce a new trait to wrap the two traits here, but it would introduce too many traits since there will be other convo, and some of the traits is likely to be changed in near future.

PR #84, opened 2026-04-10
refactor: remove FFI interfaces from double ratchets

kaichaosun:
The FFI related code here is originally for exploring how to call Rust functions from Nim. 
It's not needed anymore, as it has been clear that double ratchets will not be used via FFI directly.


osmaczko (commented):
🧹

osmaczko (approved):

jazzz (approved):
🥇

PR #85, opened 2026-04-11
feat: add flake

plopezlpz:
Add flake to be able to build from upstream in `logos-chat-module`


flake.nix
```diff
@@ -0,0 +1,91 @@
...
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    rust-overlay = {
+      url = "github:oxalica/rust-overlay";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
+  outputs = { self, nixpkgs, rust-overlay }:
+    let
+      systems = [ "aarch64-darwin" "x86_64-darwin" "aarch64-linux" "x86_64-linux" ];
```
kaichaosun, line 14: I'm wondering why windows is not supported here?

kaichaosun (approved):

flake.nix
plopezlpz: good point, I just didn't see it supported widely in other repos/modules

flake.nix
```diff
@@ -0,0 +1,91 @@
...
+            cargoLock = {
+              lockFile = ./Cargo.lock;
+              outputHashes = {
+                "chat-proto-0.1.0" = "sha256-aCl80VOIkd/GK3gnmRuFoSAvPBfeE/FKCaNlLt5AbUU=";
+              };
+            };
+
+            nativeBuildInputs = [ pkgs.perl pkgs.pkg-config pkgs.cmake ];
+            buildType = "release";
+            # Override panic=abort from workspace Cargo.toml — incompatible with buildRustPackage
```
osmaczko: Why is it incompatible? It silently undoes a deliberate safety invariant, panicking across FFI boundaries is UB.

flake.nix
```diff
@@ -0,0 +1,91 @@
...
+            CARGO_PROFILE_RELEASE_PANIC = "unwind";
+            # Tests run in CI; some require network access unavailable in the Nix sandbox
```
osmaczko: I am wondering which tests? There are no tests run against nix build afaics.

flake.nix
```diff
@@ -0,0 +1,91 @@
...
+          rustToolchain = pkgs.rust-bin.fromRustupToolchainFile ./rust_toolchain.toml;
+          rustPlatform = pkgs.makeRustPlatform {
+            cargo = rustToolchain;
+            rustc = rustToolchain;
+          };
+        in
+        {
+          default = rustPlatform.buildRustPackage {
+            pname = "libchat";
+            version = "0.1.0";
```
osmaczko:
```suggestion
            version = (builtins.fromTOML (builtins.readFile ./crates/client-ffi/Cargo.toml)).package.version;
```

osmaczko (commented):
Can you please add a CI check confirming flake is working as expected?

flake.nix
plopezlpz: I removed the unused dylib that was surfacing this problem

core/conversations/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2024"
 
 [lib]
-crate-type = ["rlib","staticlib","dylib"]
```
plopezlpz, line 7: the dylib was making the nix build fail with:

```
error: the linked panic runtime `panic_unwind` is not compiled with this crate's panic strategy `abort`
```

for now we only use `client_ffi.a` upstream so removing this

osmaczko (approved):

PR #87, opened 2026-04-16
feat: chat cli demo app

kaichaosun:
**Why this change:**
- we need a app to demo the features and its usage in libchat.
- this app tests how storage works in different platforms

**What changes:**
- terminal ui app
- local file based transport
- encrypted storage with default password

Ignore the double ratchet ffi removal, it's in https://github.com/logos-messaging/libchat/pull/84

Screenshot of cli app

chat-cli-test


bin/chat-cli/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+mod ui;
+
+use std::path::PathBuf;
+
+use anyhow::{Context, Result};
+
+/// Get the data directory (in project folder).
+fn get_data_dir() -> PathBuf {
+    // Use the directory where the binary is or current working directory
+    let manifest_dir = env!("CARGO_MANIFEST_DIR");
```
osmaczko: CARGO_MANIFEST_DIR is baked in at compile time and is never empty in practice, making the "." fallback a dead code. Please consider using simply `std::env::current_dir()` instead.

osmaczko (approved):
Nice 👍

Works for me as well

Image

bin/chat-cli/src/app.rs
```diff
@@ -0,0 +1,467 @@
+//! Chat application logic.
+
+use std::collections::HashMap;
+use std::fs;
+use std::path::PathBuf;
+
+use anyhow::{Context, Result};
+use arboard::Clipboard;
+use libchat::{ChatStorage, Context as ChatManager, Introduction, StorageConfig};
```
jazzz, line 9: If this is supposed to mimic a developer example; Rather than using Context, this should use `crates/client/src/client.rs`

bin/chat-cli/src/app.rs
```diff
@@ -0,0 +1,467 @@
...
+                if connect_parts.len() < 2 {
+                    return Ok(Some("Usage: /connect <username> <bundle>".to_string()));
+                }
+                let remote_user = connect_parts[0];
+                let bundle = connect_parts[1];
+                self.connect(remote_user, bundle)?;
+                Ok(Some(format!("Connected to {}", remote_user)))
+            }
+            "/chats" => {
+                let sessions: Vec<_> = self.state.sessions.keys().cloned().collect();
```
jazzz: Dust: Collecting this iterator doesn't immediately seem necessary. 

```
if self.state.session.is_empty() { 
...
}

for name in map.keys() {
...
}
```

bin/chat-cli/src/transport.rs
```diff
@@ -0,0 +1,143 @@
...
+        &self.user_name
+    }
+}
+
+fn now() -> u64 {
+    std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap()
+        .as_millis() as u64
+}
```
jazzz: A duplicate function exists in `app.rs`, its not immediately clear if things would break if the implementations differed. I'd consider unifying them or labelling appropriately

bin/chat-cli/README.md
```diff
@@ -0,0 +1,117 @@
+# Chat CLI
+
+A terminal chat application based on libchat library.
+
+## Features
+
+- End-to-end encrypted messaging using libchat
+- File-based transport for local simulation (no network required)
```
jazzz, line 8: Dust: Consider using Multicast instead of file-based transport. Files are inherently reliable which hides a class of usage failures.

jazzz (approved):
Overall this looks clean no blockers. I think it will be hard to keep this up to date, but as long as we are cool letting this lag a bit, then I don't see any issues.

My only notes for future consideration:
- Seems like this should be using crates/client as that is the primary interface for developers. 
- This introduces some  language which is in conflict with our existing codebase e.g. Sessions, Alice/Bob. As this resides in this repo the contents become canonical, which means we'll need to update it at some point to bring it in line with the project.

bin/chat-cli/src/app.rs
kaichaosun: Good point, will figure out the latest client crate status and how to integrate with it in new PR.

bin/chat-cli/src/app.rs
kaichaosun: Tried this, but got immutable / mutable borrow conflict in for loop.

bin/chat-cli/README.md
kaichaosun: Multicast sounds a good idea, I remembered tried it before, but switch to file based transport for some reason. Maybe still worth to try it later. 
Make the transport switchable looks cool.

kaichaosun:
> This introduces some language which is in conflict with our existing codebase e.g. Sessions, Alice/Bob. 

- I have renamed sessions to chats. 
- Alice/bob sounds proper solution here, I see there is a spec for participants naming conventions, and i have list my concern there. Overall, alice/bob is the go to names in many scenarios, and I believe this is what most developers expects in this cli demo app.
- Context is alias as ChatManager in chat-cli which is much more clear in its usage, I hope we can remove/rename Context soon.

bin/chat-cli/src/app.rs
jazzz: Ah I see. The state is being mutated while being looped over - I missed that.

PR #91, opened 2026-04-17
feat(chat-cli): wire up logos-delivery transport and switch to client API

osmaczko:
Replace the direct use of `conversations::Context` with `client::ChatClient`,
which is the intended public API for library consumers.

Remove `MessageEnvelope` and the username-keyed session model. The envelope
was never part of the wire protocol — sender identity was only tracked in the
CLI's local state. Chats are now keyed by conversation ID; add `/nickname` as
the user-facing replacement for named sessions.

Add a logos-delivery (Waku) transport alongside the existing file transport.
The active transport is selected at compile time: set `LOGOS_DELIVERY_LIB_DIR`
to link liblogosdelivery, otherwise the file transport is used.

Add logos-delivery as a Nix flake input and expose `.#logos-delivery` so the
library can be built with `nix build` and referenced by `LOGOS_DELIVERY_LIB_DIR`.

CI: rename `c-ffi-smoketest` to `smoketest`; add logos-delivery build step
and a `--smoketest` invocation of chat-cli to verify startup.


bin/chat-cli/src/transport/logos_delivery.rs
jazzz, line 1: This Project follow the Rust 2018 conventions and uses "New Style Modules" rather than "Mod" files.

https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-modrs

README.md
```diff
@@ -1,2 +1,19 @@
...
+
+## Example app
+
+[`bin/chat-cli`](bin/chat-cli/) is an end-to-end encrypted CLI chat app
+built on this library.  It uses [logos-delivery](https://github.com/logos-messaging/logos-delivery)
+(Waku-based) as the transport so two users anywhere in the world can chat by
+sharing an intro bundle.
+
+```sh
+# Build logos-delivery with Nix, then build chat-cli:
```
kaichaosun:
```suggestion
# Build logos-delivery with Nix (don't omit suffix `#logos-delivery`)
```

README.md
```diff
@@ -1,2 +1,19 @@
...
+nix build .#logos-delivery
+LOGOS_DELIVERY_LIB_DIR=./result/lib cargo build --release -p chat-cli
```
kaichaosun, line 15:
```suggestion
# Build chat-cli
LOGOS_DELIVERY_LIB_DIR=./result/lib cargo build --release -p chat-cli
```

README.md
```diff
@@ -1,2 +1,19 @@
...
+## Example app
+
+[`bin/chat-cli`](bin/chat-cli/) is an end-to-end encrypted CLI chat app
+built on this library.  It uses [logos-delivery](https://github.com/logos-messaging/logos-delivery)
+(Waku-based) as the transport so two users anywhere in the world can chat by
+sharing an intro bundle.
+
+```sh
+# Build logos-delivery with Nix, then build chat-cli:
+nix build .#logos-delivery
```
kaichaosun, line 13: I got compile error to find librln by running this command on macos,
```
error: builder for '/nix/store/n2qaffblhp6fkik1z0b4bs8h9an9gq8n-logos-delivery-patched.drv' failed with exit code 1;
       last 9 log lines:
       > Running phase: unpackPhase
       > unpacking source archive /nix/store/y2wvypx1y7m3xlzxy30bzhza64iyghq8-liblogosdelivery-dev
       > source root is liblogosdelivery-dev
       > Running phase: patchPhase
       > Running phase: updateAutotoolsGnuConfigScriptsPhase
       > Running phase: configurePhase
       > no configure script, doing nothing
       > Running phase: installPhase
       > cp: cannot stat '/nix/store/y2wvypx1y7m3xlzxy30bzhza64iyghq8-liblogosdelivery-dev/lib/librln.dylib': No such file or directory
       For full logs, run 'nix log /nix/store/n2qaffblhp6fkik1z0b4bs8h9an9gq8n-logos-delivery-patched.drv'
```

bin/chat-cli/Cargo.toml
```diff
@@ -9,10 +9,19 @@ path = "src/main.rs"
 
 [dependencies]
 libchat = { path = "../../core/conversations" }
```
kaichaosun: We can remove libchat if it's not used anymore.

bin/chat-cli/src/transport/logos_delivery/sys.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+pub fn get_trampoline<C>(_: &C) -> FFICallBack
+where
+    C: FnMut(i32, &str),
+{
+    trampoline::<C>
+}
+
+pub const RET_OK: i32 = 0;
```
kaichaosun: Better move the const values to the top of file.

bin/chat-cli/src/transport/logos_delivery.rs
```diff
@@ -0,0 +1,255 @@
...
+    message_json: String,
+    reply: mpsc::SyncSender<Result<(), DeliveryError>>,
+}
+
+type SubscriberList = Arc<Mutex<Vec<mpsc::SyncSender<Vec<u8>>>>>;
+
+// ── Config ───────────────────────────────────────────────────────────────────
+
+#[derive(Debug, Clone)]
+pub struct LogosConfig {
```
kaichaosun: Should be LogosDeliveryConfig, or maybe just use Config, since the module is already logos_delivery.

bin/chat-cli/src/transport/logos_delivery.rs
```diff
@@ -0,0 +1,255 @@
...
+
+use base64::Engine;
+use base64::engine::general_purpose::STANDARD as BASE64;
+use client::{AddressedEnvelope, DeliveryService};
+use tracing::{error, info, warn};
+
+use wrapper::LogosNodeCtx;
+
+pub fn content_topic_for(delivery_address: &str) -> String {
+    format!("/libchat/1/{delivery_address}/proto")
```
kaichaosun: The content could be `/chat-cli/private/1/...`, to help identify app, protocol, etc.

bin/chat-cli/src/transport/logos_delivery.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+impl DeliveryService for LogosDeliveryService {
+    type Error = DeliveryError;
+
+    fn publish(&mut self, envelope: AddressedEnvelope) -> Result<(), DeliveryError> {
+        let content_topic = content_topic_for(&envelope.delivery_address);
+        let payload_b64 = BASE64.encode(&envelope.data);
+
+        let message_json = serde_json::json!({
```
kaichaosun: Would be good to have struct for this fields for encoding and decoding.

bin/chat-cli/build.rs
```diff
@@ -0,0 +1,20 @@
+fn main() {
+    println!("cargo::rustc-check-cfg=cfg(logos_delivery)");
+    println!("cargo:rerun-if-env-changed=LOGOS_DELIVERY_LIB_DIR");
+
+    let Ok(lib_dir) = std::env::var("LOGOS_DELIVERY_LIB_DIR") else {
+        return;
+    };
+
+    println!("cargo:rustc-cfg=logos_delivery");
```
kaichaosun, line 9: I think the feature flag for the app is over kill, we can just enable file transport and logos_delivery, and with a param (like `--transport`) to decide which transport to use.

LOGOS_DELIVERY_LIB_DIR could use the default ones without explicitly set it when set the logos delivery params, and LOGOS_DELIVERY_LIB_DIR can be configurable if need a different path.

README.md
osmaczko: Should be resolved now, please try again

bin/chat-cli/src/transport/logos_delivery.rs
osmaczko: Good suggestion, but it does not fit the content topic specification: https://docs.waku.org/learn/concepts/content-topics. Renamed to `/logos-chat-cli/1/{delivery_address}/proto`

jazzz: > /logos-chat-cli/1/{delivery_address}/proto

This is an area where we will want to break from the spec until it can be updated.



> /{application-name}/{version}/{content-topic-name}/{encoding}
1) Choosing specific "application" name, breaks interoperability. To enable cross communication between applications, `application-name` should be generic to the service eg. `logos-chat`.

jazzz: 2) > The content could be /chat-cli/private/1/..., to help identify app, protocol, etc.

The issue is that we do not want to separate traffic by conversation protocol. They are purposely multiplexed together using `EnvelopeV1` which allows client observability into unsupported types. E.g InboxV1 - a single place to listen for all messages.

bin/chat-cli/src/main.rs
```diff
@@ -1,93 +1,189 @@
...
+    #[arg(long, default_value_t = 60000)]
+    port: u16,
+
+    /// Write logs to a file instead of stderr (keeps TUI output clean).
+    #[arg(long)]
+    log_file: Option<PathBuf>,
+
+    /// Initialize and immediately exit without launching the TUI (for CI).
+    #[arg(long)]
+    smoketest: bool,
```
jazzz, line 45: [Dust] This flag seems out of place to me. Naively appears the most value comes from ensuring the build succeeds - if the application returns early.

jazzz (approved):
I've tested it, looks good for a dev app.

bin/chat-cli/build.rs
osmaczko: I agree, let me address it in the followup PR.

kaichaosun (approved):
💯

PR #92, opened 2026-04-24
Add GroupV1 + InboxV2

jazzz:
This PR Introduces MLS Groupchat via GroupV1Convo

Changes:

- Remove "bin/chat-cli" from default members.
Having to update Applications when pushing changes for Core Protocol slows down development. Recommend moving to a separate Repo, or lock  which can lock to a specific Version. 

- Moves DeliveryService trait from Client into Core::Context. Adds RegistrationService as well. 

- Adds InboxV2 which is PQresistant.

- Adds GroupConvo Trait which extends Convo with Group Management Operation. 

- Adds GroupV1Convo which is Service aware - They can interact directly with the DeliveryService and RegistrationService impls. This streamlines the return types from Conversation functions.

- Adds `fn load_group_convo` and `fn create_group_convo` which return `Box<dyn GroupConvo>`. This allows clients to directly invoke methods and avoid relaying function calls through Context

Work Deferred:

- Any Changes to InboxV1.
- Renaming Context. Context is no longer an accurate name, however that can be discussed later.
- Client Api - This change set installs mls into core. Future work will be needed to make it usable and developer friendly.


bin/chat-cli/Cargo.toml
```diff
@@ -8,7 +8,8 @@ name = "chat-cli"
 path = "src/main.rs"
 
 [dependencies]
-client = { path = "../../crates/client" }
+# Reference a specific commit so updates to the Core does not break examples
+client = { git = "https://github.com/logos-messaging/libchat", rev = "39bf26756448dd16ddff89a6c0054f79236494aa" }
```
jazzz: This is causing NIX build to crash. 

Chat-CLI requires changes however I'd prefer not to deal with that in this PR.

core/conversations/src/conversation.rs
```diff
@@ -36,6 +42,14 @@ pub trait Convo: Id + Debug {
...
 }
 
+pub trait GroupConvo<DS: DeliveryService, RS: KeyPackageProvider>: Convo {
+    fn add_member(&mut self, members: &[&AccountId]) -> Result<(), ChatError>;
+
+    // This is intended to replace `send_message`. The trait change is that it automatically
+    // sends the payload directly.
+    fn send_content(&mut self, content: &[u8]) -> Result<(), ChatError>;
+}
+
```
jazzz, line 52: Group chats require extra functionality, compared to 1:1 Chats. 

This trait extends trait:Convo to include group management functions, while also requiring Groupchats to implement all the same functions as the other conversations

core/conversations/src/inbox_v2.rs
```diff
@@ -0,0 +1,251 @@
...
+        &self.provider
+    }
+
+    fn invite_user<DS: DeliveryService>(
+        &self,
+        ds: &mut DS,
+        account_id: &AccountId,
+        welcome: &MlsMessageOut,
+    ) -> Result<(), ChatError> {
+        let invite = GroupV1HeavyInvite {
```
kaichaosun, line 44: The version number in file name and struct name is poping up everywhere, consider following options:
- not introduce version identifier in file name.
- use enums for version variants.

core/conversations/src/inbox_v2.rs
```diff
@@ -0,0 +1,251 @@
...
+            welcome_bytes: welcome.to_bytes()?,
```
kaichaosun, line 45: Better call it message or welcome_message.

core/conversations/src/inbox_v2.rs
```diff
@@ -0,0 +1,251 @@
...
+        };
+
+        let envelope = EnvelopeV1 {
+            conversation_hint: ProtocolParams::conversation_id_for_account_id(account_id),
+            salt: 0,
+            payload: frame.encode_to_vec().into(),
+        };
+
+        let outbound_msg = AddressedEnvelope {
+            delivery_address: ProtocolParams::delivery_address_for_account_id(account_id),
```
kaichaosun: `delivery_address_for_account_id(account_id)` looks a bit long, extract deliver address conversion somewhere else (like in the scope of DeliveryAddress) could make it clear. Same for `conversation_id_for_account_id`.

core/conversations/src/inbox_v2.rs
```diff
@@ -0,0 +1,251 @@
...
+    fn delivery_address_for_account_id(account_id: &AccountId) -> String {
+        blake2b_hex::<hash_size::AccountId>(&["InboxV2|", "delivery_address|", account_id.as_str()])
+    }
+
+    fn conversation_id_for_account_id(account_id: &AccountId) -> String {
+        blake2b_hex::<hash_size::Testing>(&["InboxV2|", "conversation_id|", account_id.as_str()])
+    }
+}
+
+type ProtocolParams = InboxProtocolParams;
```
kaichaosun: Wondering why introduce an alias here?

core/conversations/src/inbox_v2.rs
```diff
@@ -0,0 +1,251 @@
...
+/// such as MLS.
+pub struct InboxV2<DS, RS, CS> {
+    account_id: AccountId,
+    ds: Rc<RefCell<DS>>,
+    reg_service: Rc<RefCell<RS>>,
+    store: Rc<RefCell<CS>>,
+    ctx: Rc<RefCell<PqMlsContext>>,
+}
+
+impl<DS, CS, RS> InboxV2<DS, RS, CS>
```
kaichaosun, line 88: By looking into the implementation, I feels this is specific for group chat based on MLS, there is no need for a generic name like `InboxV2`, it could be `MlsInbox` or similar. 
The version number here looks confused with InboxV1, before we have a multi version architecture, we should focus on concrete concept for naming.

core/conversations/src/utils.rs
```diff
@@ -6,3 +6,67 @@ pub fn timestamp_millis() -> i64 {
...
+                pub struct $name;
+                impl HashLen for $name { type Size = $size; }
+            )*
+        };
+    }
+
+    use blake2::digest::consts::{U4, U8};
+    hash_sizes! {
+        /// Generic hash size for tests and debug
+        Testing  => U4,
```
kaichaosun: Such code should be lived in tests.
Also consider not use macro if not too many boilerplate code.

core/conversations/src/utils.rs
```diff
@@ -6,3 +6,67 @@ pub fn timestamp_millis() -> i64 {
...
+        /// Account ID hash length
+        AccountId => U8,
+    }
+}
+
+use blake2::{Blake2b, Digest};
```
kaichaosun: Import should be at the top of the source file.

core/storage/src/store.rs
```diff
@@ -120,6 +123,8 @@ pub trait RatchetStore {
     fn cleanup_old_skipped_keys(&mut self, max_age_secs: i64) -> Result<usize, StorageError>;
 }
 
+// TODO: (P2) this should be defined in the ConversationType
```
kaichaosun, line 126: There are many todo items in the PR, it should be better tracked in the issue with pending tasks, not in the committed code, as there is not enough context to understand the todo item, so it's not helpful for code readers.

extensions/components/src/storage/in_memory_store.rs
```diff
@@ -0,0 +1,136 @@
...
+    // TODO: (P4) Importable crates need to be prefixed with a project name to avoid conflicts
+    ConversationMeta,
+    ConversationStore,
+    EphemeralKeyStore,
+    IdentityStore,
+    RatchetStore,
+};
+
+/// An Test focused StorageService which holds data in a hashmap
+pub struct MemStore {
```
kaichaosun, line 13: The sqlite storage should already provide in memory store, not sure if we still need this one.

core/conversations/src/inbox_v2.rs
```diff
@@ -0,0 +1,251 @@
...
+use crate::ChatError;
+use crate::DeliveryService;
+use crate::RegistrationService;
+use crate::account::LogosAccount;
+use crate::conversation::GroupConvo;
+use crate::conversation::group_v1::MlsContext;
+use crate::conversation::{GroupV1Convo, IdentityProvider};
+use crate::types::AccountId;
+use crate::utils::{blake2b_hex, hash_size};
+pub struct PqMlsContext {
```
kaichaosun, line 22: We can just flatten the fields inside this context to the above struct which is `InboxV2`, since it's not used anywhere else.

core/conversations/src/errors.rs
```diff
@@ -26,6 +27,23 @@ pub enum ChatError {
...
+    #[error("KeyPackage: {0}")]
+    KeyPackage(#[from] openmls::prelude::KeyPackageVerifyError),
+    #[error("Delivery: {0}")]
+    Delivery(String),
+}
+
+impl ChatError {
+    // This is a stopgap until there is a proper error system in place
+    pub fn generic(e: impl ToString) -> Self {
+        Self::Generic(e.to_string())
```
kaichaosun, line 45: I believe the practice here is to use Unknown(err) for such catchall error. The comment can be removed too with unknown error and implement `From<String>` trait.

core/conversations/src/context.rs
```diff
@@ -47,43 +65,97 @@ impl<S: ChatStore> Context<S> {
         let identity = Rc::new(identity);
         let inbox = Inbox::new(Rc::clone(&store), Rc::clone(&identity));
 
+        let pq_inbox = InboxV2::new(
+            LogosAccount::new_test(name),
```
kaichaosun, line 69: Raise it again, test identifiers should not made into product code, even it's temporary. It's a bad smell in code. Consider replace it with a meaningful comment.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -0,0 +1,426 @@
...
+
+use crate::types::AccountId;
+use crate::{
+    DeliveryService,
+    conversation::{ChatError, ConversationId, Convo, GroupConvo, Id},
+    service_traits::KeyPackageProvider,
+    types::{AddressedEncryptedPayload, ContentData},
+};
+
+pub trait IdentityProvider: OpenMlsSigner {
```
kaichaosun, line 31: Is this going to be shared with others or only group_v1? Since it's re exported in conversations module, need to make it clear about the scope of this trait.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -0,0 +1,426 @@
...
+        let mls_message_out = self
+            .mls_group
+            .create_message(provider, ctx_ref.ident(), content)
+            .unwrap();
+
+        let a = AddressedEncryptedPayload {
+            delivery_address: self.delivery_address(),
+            data: EncryptedPayload {
+                encryption: Some(
+                    chat_proto::logoschat::encryption::encrypted_payload::Encryption::Plaintext(
```
kaichaosun: This looks too long, partial of the uri can be put in import line.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -0,0 +1,426 @@
...
+        };
+
+        Ok(vec![a])
+    }
+
+    fn handle_frame(
+        &mut self,
+        encoded_payload: EncryptedPayload,
+    ) -> Result<Option<ContentData>, ChatError> {
+        use chat_proto::logoschat::encryption::encrypted_payload::Encryption;
```
kaichaosun: Same here, use should be in top of file.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -0,0 +1,426 @@
...
+                data: msg.into_bytes(),
+                is_new_convo: false,
+            })),
+            ProcessedMessageContent::StagedCommitMessage(commit) => {
+                self.mls_group
+                    .merge_staged_commit(provider, *commit)
+                    .map_err(ChatError::generic)?;
+                Ok(None)
+            }
+            x => {
```
kaichaosun:
```suggestion
            other => {
```

core/conversations/src/conversation/group_v1.rs
```diff
@@ -0,0 +1,426 @@
...
+}
+
+impl<MlsCtx, DS, KP> GroupConvo<DS, KP> for GroupV1Convo<MlsCtx, DS, KP>
+where
+    MlsCtx: MlsContext,
+    DS: DeliveryService,
+    KP: KeyPackageProvider,
+{
+    fn add_member(&mut self, members: &[&AccountId]) -> Result<(), ChatError> {
+        // add_members returns:
```
kaichaosun: Comment should be at the front of function.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -0,0 +1,426 @@
...
+            return Err(ChatError::Protocol(
+                "Cannot add more than 50 Members at a time".into(),
+            ));
+        }
+
+        // Get the Keypacakages and transpose any errors.
+        // The account_id is kept so invites can be addressed properly
+        let keypkgs = members
+            .iter()
+            // .map(|ident| self.key_package_for_account(ctx, ident))
```
kaichaosun: Remove it if not needed.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -0,0 +1,426 @@
...
+    service_traits::KeyPackageProvider,
+    types::{AddressedEncryptedPayload, ContentData},
+};
+
+pub trait IdentityProvider: OpenMlsSigner {
+    fn friendly_name(&self) -> String;
+    fn public_key(&self) -> &Ed25519VerifyingKey;
+}
+
+pub trait MlsContext {
```
kaichaosun, line 45: Consider flattening this if possible and avoiding the Context naming. Context can be a very overloaded abstraction and sometimes hides what the trait actually represents. A more specific name may make the API clearer.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -0,0 +1,426 @@
...
+            ctx,
+            ds,
+            keypkg_provider,
+            mls_group,
+            convo_id,
+        })
+    }
+
+    // Constructs a new conversation upon receiving a MlsWelcome message.
+    pub fn new_from_welcome(
```
kaichaosun, line 126: Is `join` fit better here?
Similarly `create` looks better for `new`.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -0,0 +1,426 @@
...
+                .build()
+                .unwrap()
+                .into_group(provider)
+                .unwrap()
+        };
+
+        let convo_id = hex::encode(mls_group.group_id().as_slice());
+        Self::subscribe(&mut *ds.borrow_mut(), &convo_id)?;
+
+        println!(
```
kaichaosun: Consider using a library for log.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -0,0 +1,426 @@
...
+        ctx: Rc<RefCell<MlsCtx>>,
+        ds: Rc<RefCell<DS>>,
+        keypkg_provider: Rc<RefCell<KP>>,
+        convo_id: String,
+        group_id: GroupId,
+    ) -> Result<Self, ChatError> {
+        let Some(mls_group) = MlsGroup::load(ctx.borrow().provider().storage(), &group_id)
+            .map_err(ChatError::generic)?
+        else {
+            return Err(ChatError::NoConvo("mls group not found".into()));
```
kaichaosun: `ok_or_else` on option can inline the logic with functional style.

core/conversations/src/inbox_v2.rs
jazzz: >  I feels this is specific for group chat based on MLS,

It is not.

InboxV2 is a breaking change to InboxV1, which offers different framing, and account support.

```
/// An PQ focused Conversation initializer.
/// InboxV2 Incorporates an Account based identity system to support PQ based conversation protocols
/// such as MLS.
```

core/conversations/src/utils.rs
jazzz: Testing removed

core/conversations/src/utils.rs
jazzz: done

core/storage/src/store.rs
jazzz: Yes all TODOs must have a corresponding issue.

extensions/components/src/storage/in_memory_store.rs
jazzz: Currently there are limitations of the ChatStore trait which adds friction.

Th largest being It combines the registration of conversations, with the storage of the actual conversation state.

This means that in order to use the get/load functionality in Context, a full SQLlite implementation needs to be provided for new conversationTypes. For PRs which are adding new functionality to `Core` this is a burden. 

The Added MemStore impl provides the lightest requirements  (Put it in a hashmap) so testing/validation can continue. 

Followup work can add persistence, and update the existing Components to support the new ConversationType

core/conversations/src/inbox_v2.rs
jazzz: > use enums for version variants.

This is a Protobuf message constructor which represents a specific Invite for a specific ConversationType. I'm not sure how you'd want to use an enum here?

core/conversations/src/inbox_v2.rs
jazzz: I don't follow; You are suggesting to rename the proto_buf field to "message"?

core/conversations/src/inbox_v2.rs
jazzz: > Extract deliver address conversion somewhere else (like in the scope of DeliveryAddress) 

DeliveryAddress generation is determined by each protocol. InboxV1, InboxV2, GroupV1 all generate delivery addresses differently. 

Moving the definition into the generic container DeliveryAddress would defeat the purpose of having multiple versioned protocols.

> delivery_address_for_account_id(account_id) looks a bit long, 

I can shorten it by some characters if that will help.

core/conversations/src/inbox_v2.rs
jazzz: removed

core/conversations/src/conversation/group_v1.rs
jazzz: Not sure I agree: 'new_' is the convention for constructing new instances of a type. Join would imply an operation on a GroupV1Convo rather than creating a new instance.

core/conversations/src/conversation/group_v1.rs
jazzz: removed

core/conversations/src/conversation/group_v1.rs
jazzz: updated to: 
```
        let mls_group = MlsGroup::load(ctx.borrow().provider().storage(), &group_id)
            .map_err(ChatError::generic)?
            .ok_or_else(|| ChatError::NoConvo("mls group not found".into()))?;
```

core/conversations/src/errors.rs
jazzz: The use of a named constructor allows this function to be passed as a `Fn`, `FnOnce` reducing complexity at callsites: e.g. result.map_err(ChatError::generic).

core/conversations/src/conversation/group_v1.rs
jazzz: > Is this going to be shared with others or only group_v1?

Not exactly sure what we mean by `shared.` 
IdentityProvider is an interface that will need to be implemented by some other code, so its required to be accessible by them.

Other conversation types in the future will likely use this trait - the conversation that supports DeMLS for example will likely reuse this trait. 

> need to make it clear about the scope of this trait.

The definition is owned by GroupV1, and accessible by objects with access to conversations internals. Comment added.

core/conversations/src/conversation/group_v1.rs
jazzz: Shortened to `encrypted_payload::Encryption::Plaintext`. This provides the necessary context, while making it easier to follow

core/conversations/src/conversation/group_v1.rs
jazzz: replace with `_` as it is unused now

core/conversations/src/inbox_v2.rs
jazzz: I see what you are saying, and don't disagree here.

However this is a ContextObject - Its expected to grow. If DeMLS is integrated and the MLSContext is still relatively simple than I could see flattening it being a smart choice.

core/conversations/src/context.rs
jazzz: > test identifiers should not made into product code
> It's a bad smell in code.

I completely agree.

> should not made into product code, even it's temporary.

I don't consider this production code at the moment.  Specifically the "new_test" is intended to warn others that this code is not to be used, and the function will be removed quickly.

> Consider replace it 
To resolve this issue is going to make this PR significantly larger, but I'm happy to do that if desired.

core/conversations/src/conversation/group_v1.rs
jazzz: > Context can be a very overloaded abstraction and sometimes hides what the trait actually represents.

Unfortunately the design pattern being used in this case is the 'Context Object Pattern'. The name specifically refers to that. I can rename it, but it would be less clear about the intention. 

> Consider flattening

I'm not sure how you would like to flatten a inter-module interface? The trait defines required functionality that needs to be supplied by the caller. I'm not sure how to remove a layer of nesting, when there is only one layer

>  A more specific name

Its the Context for all Mls Based chats. Alternatives I can see:
- `GroupV1MlsContext`: Reflects  that its currently only used in GroupV1
- `MlsContextObject`: This seems worse, as its not an object its a trait.

Do either of these align with what you are looking for

bin/chat-cli/Cargo.toml
jazzz: Rather than this being a blocker, i'm inclined to revert this and find a solution for Chat-CLI in a different PR. 

Naively seems to me that the CLI app should be put into its own repo so:
- It can properly pinned to a dependency versions
- Easily forkable / readable by developers
- Not block future development. 
- Maintain its own github issues.

core/conversations/src/inbox_v2.rs
kaichaosun: For protobuf message constructor, It can be namespaced like `v1::GroupInvite`.
Actually most of my concern comes from the version identifier splits in the file names and many other structs.

core/conversations/src/inbox_v2.rs
kaichaosun: Yes, no need for type `bytes`, `GroupV1HeavyInvite -> message` seems express the intention pretty clear.

core/conversations/src/inbox_v2.rs
kaichaosun: > DeliveryAddress generation is determined by each protocol. InboxV1, InboxV2, GroupV1 all generate delivery addresses differently.

Introduce trait and implement by different conversation should solve it clearly.

core/conversations/src/inbox_v2.rs
kaichaosun: Does `InboxV2` also works for `PrivateV1` in the current implementation or it's only for MLS?
If it's only MLS, make it clear that it's for MLSInbox, then after introduce adapter for PrivateV1, the generic `Inbox` can be introduced and delegate to the PrivateV1Inbox and MLSInbox.

core/conversations/src/inbox_v2.rs
kaichaosun: By quickly looking into the context object pattern, it looks useful to provide traced info going though the call stack like request id, or other similar shared state.
In our codebase, there are many context objects, generally I don't see strong requirement for so many context.

core/conversations/src/context.rs
kaichaosun: I mean "product" code not production, which means the logic for features, specifically not for test.

> Specifically the "new_test" is intended to warn others that this code is not to be used

A clear comment serve this purpose better, not a `*test` identifier.

core/conversations/src/conversation/group_v1.rs
kaichaosun: If owned by GroupV1, it should be `GroupIdentityProvider`, if shared by others, consider move it outside of group_v1.rs, and reduce constraint of `OpenMlsSigner` as MLS related type is specifically for group chat.

core/conversations/src/conversation/group_v1.rs
kaichaosun: I may not make myself clear.
I mean you are consolidate a lot things in the context trait, it should be split into different ones for the specific feature it provides, for example a identity provider trait could be used, not putting it in a generic context.

So instead have a big trait for a lot of things, there are many specific traits for the feature it provides. Or if you think the features are related and want to consolidate, a better name needed to express the relation / consolidatation, not a generic context, which has been over used in my opinion.

core/conversations/src/inbox_v2.rs
jazzz: > Yes, no need for type bytes

I don't agree here. The protobuf does not accept any arbitrary message. It specifically requires a Mls::Welcome   that has been encoded into bytes. Calling it a welcome message would imply that the type is a Welcome Protobuf message - where actually it is encoded bytes.

core/conversations/src/inbox_v2.rs
jazzz: I can't say I follow your logic, perhaps you could explain more.

> extract deliver address conversion somewhere else (like in the scope of DeliveryAddress) could make it clear.

The stated goal here is clarity.

> Introduce trait and implement by different conversation should solve it clearly.

A trait is an interface that exposes functionality. Its a mechanism to define an interface between components.
In this case the delivery address is never used outside of the environment it was defined.

I don't understand how exposing an internal variable or function outside of its natural scope helps with making things clear?

jazzz: I can flatten it, into a non member function and make the name less long. Hopefully that can make things more clear.

core/conversations/src/inbox_v2.rs
jazzz: > Does InboxV2 also works for PrivateV1

No it does not.  PrivateV1 does not use PQE or LogosAccounts.

> or it's only for MLS?

No it is not. It would support any any conversationType that is intended to use LogosAccounts.

core/conversations/src/inbox_v2.rs
jazzz: updated in df35abe19ad99850204bd452d6e2da6c3a99a2df

core/conversations/src/inbox_v2.rs
jazzz: > I don't see strong requirement for so many context.

There are currently 2 Contexts and soon to be 1 Context:

1) Conversations::Context which was an FFI Context pattern. This object has since changed, and as noted in my PR description need to be removed. 
 > Renaming Context. Context is no longer an accurate name, however that can be discussed later.

This PR is already large/contentious; seems like a bad idea to combine with more changes. If It will help get this merged, i'm happy to add that change set to this PR. 

2) MlsContext Trait + PqMlsContext impl 
Which is a context pattern for passing in a growing set of execution context to Mls style Conversations. 
>  or other similar shared state.

Yes, exactly. 

> If DeMLS is integrated and the MLSContext is still relatively simple than I could see flattening it being a smart choice.

Is this not sufficient?

core/conversations/src/inbox_v2.rs
jazzz: Namespacing in general is not a terrible choice here. Something to consider. 

However the natural namespace would be `inbox` as that is the owner of all invites: that means that handling inbound messages will result in a match statement which is all `GroupInvite`s. separated by artificial namespaces. 

Using:
`inbox::<ConversationTypeName>::Invite` is one approach but I'm not sure its improved.

eg:
```
inbox::group1::Invite
inbox::group2::Invite
inbox::private2::Invite
```
Flat names would be easier to reason about, and having many items all named the same seems like adding more confusion than clarity

jazzz: > Actually most of my concern comes from the version identifier splits in the file names and many other structs.

Seems your issue is with the ConversationType naming policy, not the Implementation.
ref: https://github.com/logos-co/logos-lips/pull/323

core/storage/src/store.rs
```diff
@@ -27,13 +27,15 @@ pub trait EphemeralKeyStore {
 pub enum ConversationKind {
     PrivateV1,
     Unknown(String),
```
osmaczko, line 29: Can we keep `Unknown` either first or last?

core/conversations/src/service_traits.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/// Manages key bundle storage for MLS group creation/addition while contacts are
+/// offline.
+///
+/// Implement this to provide a contact registry — ach participant publishes their key package
+/// on registration; others fetch it to initiate a conversation.
+pub trait RegistrationService: Debug {
+    type Error: Display;
+    fn register(&mut self, identity: &str, key_bundle: Vec<u8>) -> Result<(), Self::Error>;
```
osmaczko, line 25:
```suggestion
    fn register(&mut self, identity: &AccountId, key_bundle: Vec<u8>) -> Result<(), Self::Error>;
```

core/conversations/src/utils.rs
```diff
@@ -6,3 +7,66 @@ pub fn timestamp_millis() -> i64 {
...
+    /// This macro generates HashLen for the given typenum::length
+    macro_rules! hash_sizes {
+        ($($(#[$attr:meta])* $name:ident => $size:ty),* $(,)?) => {
+            $(
+                $(#[$attr])*
+                pub struct $name;
+                impl HashLen for $name { type Size = $size; }
+            )*
+        };
+    }
```
osmaczko, line 35: Image

core/conversations/src/conversation.rs
```diff
@@ -36,6 +42,14 @@ pub trait Convo: Id + Debug {
     fn convo_type(&self) -> ConversationKind;
 }
 
+pub trait GroupConvo<DS: DeliveryService, RS: KeyPackageProvider>: Convo {
+    fn add_member(&mut self, members: &[&AccountId]) -> Result<(), ChatError>;
+
+    // This is intended to replace `send_message`. The trait change is that it automatically
```
osmaczko, line 48:
```suggestion
    // TODO: This is intended to replace `send_message`. The trait change is that it automatically
```

core/integration_tests_core/tests/mls_integration.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+fn process(clients: &mut Vec<Client>) {
+    for client in clients {
+        client.process_messages();
+    }
+}
+
+#[test]
+fn create_group() {
```
osmaczko, line 79: Would be great to add assertions to this test.

extensions/components/src/delivery/local_broadcaster.rs
```diff
@@ -0,0 +1,122 @@
...
+        self.messages.get(cursor + self.base_index)
+    }
+
+    pub fn tail(&self) -> usize {
+        self.messages.len() + self.base_index
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct LocalBroadcaster {
```
osmaczko, line 28: Should `InProcessDelivery` be removed then?

core/conversations/src/conversation/group_v1.rs
```diff
@@ -0,0 +1,417 @@
...
+};
+
+/// Provides the identity information needed to participate in an MLS group.
+///
+/// Implementors must also implement [`OpenMlsSigner`] so they can sign MLS
+/// messages. The two methods here supply what [`MlsContext::get_credential`]
+/// needs to build a [`CredentialWithKey`]: `friendly_name` becomes the
+/// `BasicCredential` label and `public_key` becomes the signature-verification key.
+pub trait IdentityProvider: OpenMlsSigner {
+    fn friendly_name(&self) -> String;
```
osmaczko, line 32:
```suggestion
    fn display_name(&self) -> String;
```

bin/chat-cli/Cargo.toml
osmaczko: I see chat-cli as a demo for this repo, and I think it's healthy to keep it buildable and runnable on the same revision as core. A few reasons:
- It's the smallest end-to-end consumer of the public API. If a core change breaks it, that's signal the public surface shifted - exactly the feedback we want at PR time, not weeks later.
- Demos are the first thing new contributors touch. A broken or git-pinned demo is a poor first impression and undermines trust in the rest of the codebase.

osmaczko (commented):
Cute. I see it as a good step forward.

bin/chat-cli/Cargo.toml
jazzz: > I see chat-cli as a demo for this repo, and I think it's healthy to keep it buildable and runnable on the same revision as core. A few reasons:

@osmaczko I agree. There should be an usable app, and it its important its up to date.

Where I disagree is with how coupled these two things are - and the harm they are causing. 

Two specific examples:
- Currently deploying a group chat interface requires me to implement UI logic to support multiple conversation members. Never mind working through an application that I am not familiar with to make it compile. 
- Adding a new conversationType requires that a database schema be created, and all storage functions/queries implemented or tests fail. 

This is need to be done before the PR has even been reviewed. Its not certain that more updates to these components will be required. 

These should be kept up to date, however they can occur after the the core PR has landed. This is especially helpful while major changes are occurring.


## Potential Solution
It feels we could allow core changes to be deployed, and avoid negative consequences with process. 

- Integration tests allow surfacing of changes to the public API's, while using a simplified types that do not contain application specific logic, Developers must show that these tests pass to be accepted: https://github.com/logos-messaging/libchat/pull/92/changes#diff-f7e99e5ced818e53f5f63b2a952461475a9b1ac25e99ddb1654ffe8bd41a9f40R79

- We use a gitflow like approach (main, dev, feature_branch) to track the work. Different developers can handle the different tasks, a requirement for merging to main is that all examples and demos function. 

- Or pin larger components to a git ref, and automatically create github issues to update the items and bring them up to date. We set shelling fences at (1 Week, to ensure) they get updated. CI checks run on intervals to detect anything lagging behind.

Pinning/Branches means that the clients are always functional, work is spread across the team and PRs don't need to be 4000 lines. Leading to more thoughtful review.

core/conversations/src/conversation/group_v1.rs
jazzz: > consider move it outside of group_v1.rs, 
This is valid.

I can move it arbitrarily to InboxV2 where it belongs, however I've often been accused of premature optimization 😆  - so I tried to keep it as local and contained as it should be now and move it when there is good reason. 

> , and reduce constraint of OpenMlsSigner as MLS related type is specifically for group chat.

These two problems have been solved for [GroupV2](https://github.com/logos-messaging/libchat/pull/96) however backporting here would increase lines by another 500 lines. 

Is that what you'd like?

core/conversations/src/conversation/group_v1.rs
jazzz: Based on your feedback I ended up renaming it MlsProvider in my working branch,  - Not sure if that helps at all.

> So instead have a big trait for a lot of things, there are many specific traits for the feature it provides.

Thats going to be hard. OpenMls accepts a 'Provider' object which has cryptography, storage into one. At somepoint those functionalities are going to need to be implemented together. As mentioned elsewhere Accounts are moved outside of the conversations crate all together which simplifies this object. But the concept still remains

core/conversations/src/errors.rs
jazzz: > I believe the practice here is to use Unknown(err) for such catchall error.
Correct. 

However there is a semantic difference. 

`Unknown` is a truly unknown type.  
`Generic` is being used as a catchall until someone tackles proper error handling. Currently short on dev power.

All Generics MUST be converted to a real type.

core/conversations/src/service_traits.rs
jazzz: Good eyes; It should definitely be AccountId

extensions/components/src/delivery/local_broadcaster.rs
jazzz: Eventually that would make sense. Or InProcess could be improved better than LocalBcaster.

I could't use InProcess because it didn't support subscriptions, which were needed.

Currently the Client requires a functional change to use Subscriptions if its going to update to LocalBcaster. 
So I chose to leave it as is, and then update it when possible.

bin/chat-cli/Cargo.toml
osmaczko: > * Currently deploying a group chat interface requires me to implement UI logic to support multiple conversation members. Never mind working through an application that I am not familiar with to make it compile.

This can be done in a follow-up step, right? The crucial part is that it builds and runs. It's fine if it keeps supporting only private convos for now.

> * Adding a new conversationType requires that a database schema be created, and all storage functions/queries implemented or tests fail.

Given the new conversation type will be covered by UI logic in a follow-up, can't we just stub the storage interface with todo!()?

> * Integration tests allow surfacing of changes to the public API's, while using a simplified types that do not contain application specific logic, Developers must show that these tests pass to be accepted: https://github.com/logos-messaging/libchat/pull/92/changes#diff-f7e99e5ced818e53f5f63b2a952461475a9b1ac25e99ddb1654ffe8bd41a9f40R79

I agree, and I believe we should have full coverage of the public API within integration tests.

> * We use a gitflow like approach (main, dev, feature_branch) to track the work. Different developers can handle the different tasks, a requirement for merging to main is that all examples and demos function.
> * Or pin larger components to a git ref, and automatically create github issues to update the items and bring them up to date. We set shelling fences at (1 Week, to ensure) they get updated. CI checks run on intervals to detect anything lagging behind.

Good ideas, but I don't like either of them. The Gitflow-like approach means dev is allowed to have a broken demo, and pinning larger components to a git ref of the  same repo feels odd as it's hard to navigate and reason about the code (inception). I really see two options:
                                                                                                                                                                  
1. Keep the demo buildable and runnable, add stubs if needed, follow up with features to keep PRs reasonably sized.                                                  
2. Move the demo to a separate repo, as you proposed before.

I'm still opting for 1).

core/integration_tests_core/src/lib.rs
```diff
@@ -0,0 +1 @@
+
```
kaichaosun, line 1: It seems not used anymore.

kaichaosun (approved):
👍  well done on the the MLS.
I don't see major blockers to move it forward, future improvements could be added later,  like version management, namings.

osmaczko (approved):

bin/chat-cli/Cargo.toml
jazzz: I've updated the CLI demo so it builds. 

The future not sure the best plan.

core/conversations/src/context.rs
jazzz: This will be immediately replaced in a follow up

PR #93, opened 2026-04-24
Add Signature + Verifying key types

jazzz:
This PR prepares for Integrating MLS Conversation type, by adding functionality to generate and verify Ed25519 Signatures.


core/crypto/src/signatures.rs
```diff
@@ -0,0 +1,83 @@
...
+        Self(bytes)
+    }
+}
+
+#[derive(Clone)]
+pub struct Ed25519SigningKey(ed25519_dalek::SigningKey);
+
+impl Ed25519SigningKey {
+    pub fn generate() -> Self {
+        Self(ed25519_dalek::SigningKey::generate(&mut OsRng))
```
kaichaosun, line 37: It would be great if we can make the crate usage consistent with other parts like `Signer`.

core/crypto/src/signatures.rs
```diff
@@ -0,0 +1,83 @@
...
+use std::fmt::Debug;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+#[error("verification failed of the Ed25519 Signature")]
+pub struct SignatureVerificationError {}
+
+/// A 64-byte XEdDSA signature over an Ed25519-compatible curve.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Ed25519Signature([u8; 64]);
```
kaichaosun, line 12: The whole file seems ed25519 related, better name the file for its purpose like `ed25519`.
Also if there needs to be an mutli signature module, better to use an enum for different signature variants.

osmaczko (approved):

core/crypto/src/signatures.rs
jazzz: I'm not sure I follow. What specifically would you like to see changed?

core/crypto/src/signatures.rs
jazzz: > Also if there needs to be an mutli signature module, better to use an enum for different signature variants.

Lets wait for additional signature types before optimizing this. 

> The whole file seems ed25519 related, better name the file for its purpose like ed25519.

Don't think I agree here, but perhaps I am missing something. From a top down approach, I'd rather the imports be `crypto::signatures::Ed25519` which allows for easier discovery of available functionality. If more signatures are ever added, ed25519 can be moved to a submodule.

PR #94, opened 2026-04-24
Add Account Struct

jazzz:
This PR Adds a LogosAccount struct to prepare for MLS and PQ identity. 

This is added separately to reduce the noise in #92 

## Notes
- Depends on #93 
- Introduces a dependency on Openmls. The dependency is needed in the crate, however the direct dependency will be removed from `Account` in the future.

### DIFF
--- a/core/conversations/src/account.rs
+++ b/core/conversations/src/account.rs
@@ -4,7 +4,7 @@ use openmls_traits::signatures::Signer;
 
 use crate::types::AccountId;
 
-/// Logos Account provides represents a single account across
+/// Logos Account represents a single account across
 /// multiple installations and services.
 pub struct LogosAccount {
     id: AccountId,
--- a/core/conversations/src/account.rs
+++ b/core/conversations/src/account.rs
@@ -13,6 +13,9 @@ pub struct LogosAccount {
 }
 
 impl LogosAccount {
+    /// Create an LogosAccount using a pre-defined identifier.
+    /// This should only be used in test scenarios where the identifiers can be chosen
+    /// to ensure no conflicts between instances. Not suitable for production use.
     pub fn new_test(explicit_id: impl Into<String>) -> Self {
         let signing_key = Ed25519SigningKey::generate();
         let verifying_key = signing_key.verifying_key();
--- a/core/conversations/src/account.rs
+++ b/core/conversations/src/account.rs
@@ -29,7 +29,7 @@ impl LogosAccount {
 }
 
 impl Signer for LogosAccount {
-    // TODO: (P2) Remove OpenMLS references
+    // TODO: (P2) Remove OpenMLS dependency to make accounts more portable
     fn sign(&self, payload: &[u8]) -> Result<Vec<u8>, openmls_traits::signatures::SignerError> {
         Ok(self.signing_key.sign(payload).as_ref().to_vec())
     }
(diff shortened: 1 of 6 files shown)


core/conversations/src/account.rs
```diff
@@ -0,0 +1,40 @@
...
+/// Logos Account provides represents a single account across
+/// multiple installations and services.
+pub struct LogosAccount {
+    id: AccountId,
+    signing_key: Ed25519SigningKey,
+    verifying_key: Ed25519VerifyingKey,
+}
+
+impl LogosAccount {
+    pub fn new_test(explicit_id: impl Into<String>) -> Self {
```
kaichaosun, line 18: Better use `new`, with comment on when to use it and the plan to move forward.

core/conversations/src/account.rs
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+
+    pub fn account_id(&self) -> &AccountId {
+        &self.id
+    }
+}
+
+impl Signer for LogosAccount {
+    // TODO: (P2) Remove OpenMLS references
```
kaichaosun: I'm confusing what this comment means.

core/conversations/src/account.rs
```diff
@@ -0,0 +1,40 @@
+use crypto::{Ed25519SigningKey, Ed25519VerifyingKey};
+use openmls::prelude::SignatureScheme;
+use openmls_traits::signatures::Signer;
+
+use crate::types::AccountId;
+
+/// Logos Account provides represents a single account across
```
osmaczko:
```suggestion
/// Logos Account represents a single account across
```

core/conversations/src/account.rs
```diff
@@ -0,0 +1,40 @@
...
+/// multiple installations and services.
+pub struct LogosAccount {
+    id: AccountId,
+    signing_key: Ed25519SigningKey,
+    verifying_key: Ed25519VerifyingKey,
```
osmaczko: Seems redundant. Isn't it cheaply derivable from `signing_key`?

osmaczko (approved):

core/conversations/src/account.rs
jazzz: Its not that expensive to generate, however it does result in an allocation for every call. Originally the struct contained no verifying_key however, all consumers require a reference to it. 

Seemed more straight forward to store the bytes, rather than worrying about ownership.

If its a sticking point happy to revert it, however it seemed less clear to me

core/conversations/src/account.rs
jazzz: New would imply that this is a constructor for a production Account. This is specifically to be used in test scenarios or clients prior to the blockchain integration. Calling it the primary constructor seems like it would be inaccurate and lead to breaking changes.

jazzz: <Added a comment to provide more clarity>

jazzz: Following up here, incase there is any confusion: given this function will be removed entirely soon - I thought including "test" in the name would be a warning for everyone - I've added a comment to reflect that as well. 

I don't have any blockers if anyone feels it needs to be changed - But my focus is entirely on decomposing #92  to make it easier to review.

kaichaosun (approved):
💯

PR #95, opened 2026-04-27
chore(chat-cli): switch transport at runtime via --transport flag

osmaczko:
Both file and logos-delivery transports are now compiled into a single binary and selected at runtime (default: logos-delivery), replacing the env-var-driven build-time cfg.

Motivation: https://github.com/logos-messaging/libchat/pull/91#discussion_r3121527329

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,8 +16,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: rustup update stable && rustup default stable
-      - run: cargo build --verbose
-      - run: cargo test --verbose
+      # chat-cli's build.rs unconditionally links liblogosdelivery and requires
+      # LOGOS_DELIVERY_LIB_DIR. The smoketest job builds and exercises it under
+      # Nix; here we keep the toolchain-only job fast by skipping it.
+      - run: cargo build --verbose --workspace --exclude chat-cli
+      - run: cargo test --verbose --workspace --exclude chat-cli
 
   clippy:
     name: Clippy
@@ -26,7 +29,7 @@ jobs:
       - uses: actions/checkout@v4
       - run: rustup update stable && rustup default stable
       - run: rustup component add clippy
-      - run: cargo clippy --all-targets --all-features -- -D warnings
+      - run: cargo clippy --all-targets --all-features --workspace --exclude chat-cli -- -D warnings
 
   fmt:
     name: Format
--- a/bin/chat-cli/src/main.rs
+++ b/bin/chat-cli/src/main.rs
@@ -3,13 +3,21 @@ mod transport;
 mod ui;
 mod utils;
 
-use std::path::PathBuf;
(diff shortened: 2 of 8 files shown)


bin/chat-cli/src/main.rs
```diff
@@ -48,30 +59,54 @@ struct Cli {
...
     std::fs::create_dir_all(&cli.data).context("failed to create data directory")?;
 
-    println!("Starting chat as '{}'...", cli.name);
-    println!("Data dir: {}", cli.data.display());
+    match cli.transport {
+        TransportKind::File => {
+            let transport_dir = cli.data.join("transport");
+            let (transport, inbound) = transport::file::FileTransport::new(&transport_dir)
+                .context("failed to create file transport")?;
+            run(transport, inbound, &cli)
```
jazzz, line 70: [Dust] Managing two separate invocations is likely to cause bugs in the future. You could return a  `Box<dyn DeliveryService>;` from the match/setup function. This means that the invocation site for all transports would be the same.

bin/chat-cli/src/main.rs
```diff
@@ -48,30 +59,54 @@ struct Cli {
...
+        }
+        TransportKind::LogosDelivery => {
+            use transport::logos_delivery::{Config, Service};
+
+            eprintln!("Starting logos-delivery node (preset={})...", cli.preset);
+            eprintln!("This may take a few seconds while connecting to the network.");
```
kaichaosun: I'm wondering why use `eprintln`, not `println` or to the log file?

kaichaosun (approved):

.github/workflows/ci.yml
```diff
@@ -26,7 +29,7 @@ jobs:
       - uses: actions/checkout@v4
       - run: rustup update stable && rustup default stable
       - run: rustup component add clippy
-      - run: cargo clippy --all-targets --all-features -- -D warnings
+      - run: cargo clippy --all-targets --all-features --workspace --exclude chat-cli -- -D warnings
```
jazzz, line 32: These  `--exclude chat-cli`'s are telling me that `chat-cli` crate likely does not belong inside of this repo if needs to keep being handled differently.

jazzz (approved):
No blockers - love it.

jazzz:
Is this waiting on anything in particular? Or is it good to merge?

bin/chat-cli/src/main.rs
osmaczko: `Box<dyn DeliveryService>` doesn't compose directly, the trait's associated `Error` differs between impls (`FileTransportError` vs `DeliveryError`), so unifying it means erasing to `Box<dyn Error>`.

.github/workflows/ci.yml
osmaczko: That's a fair point. I like to think of chat-cli as a demo that should live in this repo, although there are also reasons to move it out, especially as it matures and becomes more complex.

Issue #97, opened 2026-05-04
Client Event System


jazzz:
## Problem

Currently Applications learn that they have been invited to a new group by watching for a boolean flag set on [`ContentData`](https://github.com/logos-messaging/libchat/blob/39bf26756448dd16ddff89a6c0054f79236494aa/core/conversations/src/types.rs#L16-L20). This approach is brittle, and does not scale with added features.

Specifically: Currently it is required that a new Conversation always contains some content, or else no payload is passed back the to applications. MLS works differently and there is not guaranteed to be a content message, when a participant first joins the group.

In these cases its not clear how to pass data to applications.

## Solution

A system that allows clients to notify applications of changes and updates from the Clients state. This includes:
- new conversations
- delivery receipts
- reliability failures.

Design Criteria:
- Positive Developer Experience
- Clear growth path - New events will likely be added - helpful to have a clear plan how those would be supported.

Features:
- Events + data can be passed to applications without requiring content

jazzz:
Excited to see what your plan for this is @osmaczko.

osmaczko:
Completed by https://github.com/logos-messaging/libchat/pull/99, https://github.com/logos-messaging/libchat/pull/106, https://github.com/logos-messaging/libchat/pull/123, https://github.com/logos-messaging/libchat/pull/125

PR #99, opened 2026-05-14
chore: add adr for client-event-system

osmaczko:
Adds an ADR for the client event system (#97).

Aim: agree on the design before implementation. Discuss, iterate, then merge so the doc becomes durable context for future contributors (AI included).

### DIFF
--- a/docs/adr/0001-client-event-system.md
+++ b/docs/adr/0001-client-event-system.md
@@ -1,0 +1,315 @@
+# Client Event System
+
+| Field | Value |
+|---|---|
+| Status | Accepted |
+| Issue | https://github.com/logos-messaging/libchat/issues/97 |
+| Date | 2026-05-19 |
+
+## Context and Problem
+
+Applications currently learn about new conversations from an `is_new_convo: bool` flag on `ContentData` (`core/conversations/src/types.rs:16-20`). Two problems:
+
+1. The flag overloads `ContentData`: protocol metadata is smuggled through a content carrier.
+2. The flag assumes every new conversation carries an initial content frame. Protocols such as MLS allow a conversation to begin without one; in that case `handle_payload` returns `None` and the application never observes the new conversation.
+
+Issue #97 calls for a proper event system that can signal new conversations, delivery receipts, and reliability failures — without piggy-backing on content — and that provides a clear path for adding new event types later.
+
+This ADR specifies the layered design of the event system and how events reach the application.
+
+## Decision Drivers
+
+- **Simplicity of the core.** Fully synchronous and caller-driven: no background work, no callbacks out. External effects are performed through services injected as method parameters.
+- **Extensibility.** A new event type is a localised change (one enum variant, one emit site) that does not break existing consumers.
+- **FFI compatibility.** Must remain expressible through the existing `safer-ffi` boundary in `crates/client-ffi`. Event payloads are limited to owned, concrete data (bytes, strings, identifiers) — no closures, generics, or non-`'static` references.
+
+## Architecture
+
+The library is organised in three layers. Calls flow downward; events flow upward.
+
+```mermaid
+flowchart TB
+    A["<b>app</b><br/>UI/UX layer<br/>drives the event loop"]
+    B["<b>client</b><br/>owns services<br/>runs background threads"]
+    C["<b>core</b><br/>strict sync, caller-driven"]
+
+    A -- "method calls" --> B
+    B -- "method calls" --> C
+    C -.->|"events (from method returns)"| B
+    B -.->|"events (sync + background)"| A
+```
+
+Crates: **app** — `bin/chat-cli`, future `logos-chat-module`; **client** — `crates/client`, `crates/client-ffi`; **core** — `core/conversations` and friends in libchat.
+
+## Design
+
+### Core layer
+
+#### Constraints
+
+- Strict sync, single-threaded.
+- No background work, timers, or internal queues.
+- External effects (delivery; future registration / identity lookups) are performed through services injected as method parameters.
+
+#### Approach
+
+Methods receive the services they need and call them directly. Observations (events) are returned so the caller can surface them upward:
+
+```rust
+impl<S: ChatStore> Context<S> {
+    pub fn handle_payload<D: DeliveryService>(
+        &mut self,
+        delivery: &mut D,
+        payload: &[u8],
+    ) -> Result<Vec<Event>, ChatError>;
+
+    pub fn send_content<D: DeliveryService>(
+        &mut self,
+        delivery: &mut D,
+        convo: ConversationId,
+        content: &[u8],
+    ) -> Result<Vec<Event>, ChatError>;
+
+    pub fn create_private_convo<D: DeliveryService>(
+        &mut self,
+        delivery: &mut D,
+        intro: &Introduction,
+        content: &[u8],
+    ) -> Result<(ConversationIdOwned, Vec<Event>), ChatError>;
+}
+```
+
+### Client layer
+
+#### Responsibility split
+
+The client owns the concrete service implementations (delivery, future registration, identity), polls the transport on a background thread, and processes inbound bytes by calling into the core. The application invokes client methods and consumes events; raw transport bytes (encrypted envelopes off the wire) are handled entirely inside the client.
+
+#### Constraints
+
+- Owns the concrete service implementations and injects them into core method calls.
+- Events from synchronous calls flow through the method's return type, inherited from the core.
+- Polls the transport on a background thread and feeds inbound payloads into the core.
+- May spawn additional background threads (e.g. for timer-driven retries).
+- Background threads emit events that no caller-invoked method can return — for example `DeliveryFailed { reason: Timeout }`.
+
+#### Common shape (all options)
+
+The client invokes core methods with its services; the core publishes envelopes directly through the injected delivery service. Only events flow back as return values.
+
+```rust
+impl<D: DeliveryService> ChatClient<D> {
+    pub fn send_message(&mut self, convo: &ConversationIdOwned, content: &[u8])
+        -> Result<Vec<Event>, ClientError<D::Error>>;   // sync events from this send
+
+    // Background events (including those from inbound payload processing) reach the
+    // application through one of the three mechanisms below.
+}
+```
+
+The three options differ only in how background events reach the application.
+
+#### Option A — internal poll queue
+
+The client owns a `Mutex<VecDeque<Event>>`. Background threads push to it; the application drains via two new methods.
+
+```rust
+impl<D: DeliveryService> ChatClient<D> {
+    pub fn poll_event(&mut self) -> Option<Event>;
+    pub fn drain_events(&mut self) -> Vec<Event>;
+}
+```
+
+Prior art: mio's `Events` (per-`Poll` instance, drained by the caller); rdkafka's `Consumer::poll` (background thread fills a queue, caller polls — same domain).
+
+**Pros**
+
+- Single primitive (mutex-protected queue) with no new dependencies.
+- FFI mapping is direct: `client_poll_event` returns an opaque `Option<Event>`, mirroring the existing `PushInboundResult` shape (`crates/client-ffi/src/api.rs:49-55`).
+- Matches the existing chat-cli tick-loop consumer pattern (`bin/chat-cli/src/app.rs:144-180`).
+
+**Cons**
+
+- Requires the application to drain after every operation; events accumulate if it forgets.
+- Adds shared mutable state (`Mutex<VecDeque>`) inside the client; the queue must be bounded with explicit overflow handling.
+
+#### Option B — channel handed to the caller (selected)
+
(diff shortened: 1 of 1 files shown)


docs/adr/0001-client-event-system.md
```diff
@@ -0,0 +1,285 @@
...
+Crates: **app** — `bin/chat-cli`, future `logos-chat-module`; **client** — `crates/client`, `crates/client-ffi`; **core** — `core/conversations` and friends in libchat.
+
+## Considered Options
+
+### Core layer
+
+#### Constraints
+
+- Strict sync, single-threaded.
+- No background work, timers, or internal queues.
```
osmaczko, line 51: @jazzz does this constrain hold with de-mls integration?

docs/adr/0001-client-event-system.md
```diff
@@ -0,0 +1,285 @@
...
+| Date | 2026-05-14 |
+
+## Context and Problem
+
+Applications currently learn about new conversations from an `is_new_convo: bool` flag on `ContentData` (`core/conversations/src/types.rs:16-20`). Two problems:
+
+1. The flag overloads `ContentData`: protocol metadata is smuggled through a content carrier.
+2. The flag assumes every new conversation carries an initial content frame. Protocols such as MLS allow a conversation to begin without one; in that case `handle_payload` returns `None` and the application never observes the new conversation.
+
+Issue #97 calls for a proper event system that can signal new conversations, delivery receipts, and reliability failures — without piggy-backing on content — and that provides a clear path for adding new event types later.
```
kaichaosun, line 16: It seems there are 2 types of events,
1. the events that `libchat` gets notified when `logos-delivery` has state changes, for example, message been acked via SDS.
2. the events that `app` gets notified when `libchat` has state changes. 

Does this ADR focus both scenarios or only one of the scenarios?

`chat-client` as a glue layer,
- when between `app` and `libchat` (for scenario 2), could start with poll option.
- if between `logos-delivery` and `libchat`, it seems call back is widely used when integrate with nim-sds/nwaku.

docs/adr/0001-client-event-system.md
```diff
@@ -0,0 +1,285 @@
...
+
+This ADR proposes a layered design and presents the per-layer options.
+
+## Decision Drivers
+
+- **Simplicity of the core.** Fully synchronous and caller-driven: no background work, no callbacks out, no side effects beyond storage I/O.
+- **Extensibility.** A new event type is a localised change (one enum variant, one emit site) that does not break existing consumers.
```
jazzz, line 23: A rust Enum variant here would have breaking changes for any one who matches on it. 

Not strictly a bad thing.

Decoupling would require a struct with a u16 field and associated consts.

docs/adr/0001-client-event-system.md
```diff
@@ -0,0 +1,285 @@
...
+- **FFI compatibility.** Must remain expressible through the existing `safer-ffi` boundary in `crates/client-ffi`.
```
jazzz: I like this. 

I would clarify the requirement as " Must consist of only POD"

docs/adr/0001-client-event-system.md
jazzz: Yes.

With the caveat that DeMLS does require timers for consensus. I believe this will be handled synchronously. Core still based on direct invocations.

docs/adr/0001-client-event-system.md
```diff
@@ -0,0 +1,285 @@
...
+        -> Result<CreatePrivateConvoOutput, ChatError>;
+}
+
+pub struct HandlePayloadOutput {
+    pub events: Vec<Event>,                 // observations for the app to surface (e.g. MessageReceived)
+    pub outbound: Vec<AddressedEnvelope>,   // responses for the client to dispatch (e.g. acks)
+}
+```
+
+Returning both events and outbound envelopes from the same struct keeps every side effect of a core method visible in its return type. The client layer dispatches the envelopes and surfaces the events upward; the core itself performs no I/O beyond caller-initiated storage access.
```
jazzz: No blocker; I think this makes sense.

FYI: I will be proposing a change to how the core interacts with "external service".
Ref: https://github.com/logos-messaging/libchat/blob/6dc027124f47f0a6b783ccd17d4bc60f4a59cd67/core/core_client/src/conversation/group_v1.rs#L136-L148

In normal operation a ConversationType requires access to many different services. DeliveryService is easy as it can be deferred to the client - Envelopes are returned as you describe here. Some other services are more difficult:

- Registration Service which publishes key-bundles to server and fetches them.
- Accounts/Identity which keybundles are valid.
- Wakeup services.
- MLS Crypto/Storage Providers

Rather than separating the logic of how these services are needed, the reference link provided shows the use of a ServiceContext which is passed to every function call. It contains all the available Abstract services that a Conversation may need so it can interact directly, without complicated ownership. 

In this model the ConversationTypes can directly publish messages at the callsite without having to "bubble up".  Logic doesn't bleed outside of the Conversations and memory management is clearer. 

Theres discussion about whether this approach is viable. However it may impact the decision here.

docs/adr/0001-client-event-system.md
```diff
@@ -0,0 +1,285 @@
...
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub enum FailureReason {
+    Transport,    // synchronous transport error on publish
+    PeerRejected, // peer signalled rejection (future protocol work)
+    Timeout,      // no receipt within the retry window
+}
+```
+
+`#[non_exhaustive]` on the enum permits new variants; on each struct variant it permits new fields. Both are additive minor-release changes. Future variants (`ConversationRekeyed`, `ParticipantJoined`, `PresenceChanged`, transport health, key-rotation reminders, …) follow this rule.
```
jazzz, line 251: TIL

docs/adr/0001-client-event-system.md
```diff
@@ -0,0 +1,285 @@
...
+
+| Variant | Emitted from |
+|---|---|
+| `ConversationStarted` (responder side) | `core/conversations/src/inbox/handler.rs:155-162` (replaces `is_new_convo: true`) |
+| `MessageReceived` | `core/conversations/src/conversation/privatev1.rs:184-191` (replaces `is_new_convo: false`) |
+| `DeliveryReceipt` | `Context::handle_payload` when decoding a `PrivateV1Frame::Receipt` (future protocol work) |
+| `DeliveryFailed { Transport }` | `ChatClient::dispatch_all` (`crates/client/src/client.rs:84-92`) on `delivery.publish` error |
+| `DeliveryFailed { Timeout }` | client's background retry thread |
+
+The initiator side does not emit `ConversationStarted`: `create_conversation` returns the new `ConversationIdOwned` directly.
```
jazzz: I think this is an good distinction. 

Events are for notifying applications of "Side Effects" rather than results of direct action.

docs/adr/0001-client-event-system.md
```diff
@@ -0,0 +1,285 @@
...
+
+```rust
+impl<D: DeliveryService> ChatClient<D> {
+    pub fn receive(&mut self, payload: &[u8])
+        -> Result<ReceiveOutput, ClientError<D::Error>>;       // events from this payload
+    pub fn send_message(&mut self, convo: &ConversationIdOwned, content: &[u8])
+        -> Result<SendMessageOutput, ClientError<D::Error>>;   // sync events from this send
+
+    // Background events are delivered via one of the three mechanisms below.
+}
```
jazzz, line 107: I'm not following here:

The delivery service needs to emit ApplicationEvents?

I would have assumed:
- Client polls, or registers for callbacks. I.e is notified about new data.
- Client passes data to Core::handle_payload.

What am I missing?

osmaczko: > The delivery service needs to emit ApplicationEvents?

No. This comment is a part of the `ChatClient`. I assumed that Clients may have background threads doing some work.

> Background threads emit events that no caller-invoked method can return — for example DeliveryFailed { reason: Timeout }.

For instance, a loop that retries failed sends, etc.

Please also take a look at open question, whether Client should absorb transport polling:

> 3. Does the client absorb transport polling? Today the application drives transport polling and feeds bytes into client.receive (bin/chat-cli/src/app.rs:46-87, 144-180). The client could absorb this as an additional background thread, in which case DeliveryService would become bidirectional and the application would consume events instead of bytes. Orthogonal to the event-system shape, but reshapes the application-layer contract.

In that case, handling messages would become a background work as well, which I think makes sense.

docs/adr/0001-client-event-system.md
osmaczko: > Rather than separating the logic of how these services are needed, the reference link provided shows the use of a ServiceContext which is passed to every function call. It contains all the available Abstract services that a Conversation may need so it can interact directly, without complicated ownership.

I think `ServiceContext` makes the call site simpler but weakens readability. The method signature no longer communicates what may happen as it lets methods perform any external service operation internally. I would opt for passing only the needed services to each method.

> In this model the ConversationTypes can directly publish messages at the callsite without having to "bubble up". Logic doesn't bleed outside of the Conversations and memory management is clearer.

I think it makes sense in general.

I liked the "bubble-up" approach because of its visible effects, i.e. the method signature tells you it may produce outbound envelopes and a certain set of events. But the bubble-up model clearly degrades when an operation needs a service's result mid-method.

Moreover, in this model the "no side effects beyond storage I/O" invariant no longer holds. Service calls produce side effects, and those will be hidden inside the implementation.

Testing will be a little bit harder as well, since it will require faking/mocking the services instead of asserting on the methods' return values.

docs/adr/0001-client-event-system.md
jazzz: Ah I was miss reading. Makes sense.

docs/adr/0001-client-event-system.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Open Questions
+
+1. **Sync vs async at the client layer.** The core stays sync. The client could adopt an async runtime (e.g. `tokio`) without changing the option set, but each option's natural shape changes: Option A → `Stream` over a notify primitive; Option B → `tokio::sync::mpsc::Receiver<Event>` with an `impl Stream` shape; Option C → `async fn` callback.
+
+2. **Consumer pattern assumed.** Different consumer archetypes favour different shapes: a polling UI loop suits Option A; a worker thread that blocks on `recv` suits Option B; a low-latency or push-driven consumer (toast notifications, daemons) suits Option C. Pick one — supporting multiple shapes is a maintenance burden.
+
+3. **Does the client absorb transport polling?** Today the application drives transport polling and feeds bytes into `client.receive` (`bin/chat-cli/src/app.rs:46-87, 144-180`). The client could absorb this as an additional background thread, in which case `DeliveryService` would become bidirectional and the application would consume events instead of bytes. Orthogonal to the event-system shape, but reshapes the application-layer contract.
```
jazzz: >  The client could absorb this as an additional background thread, in which case DeliveryService would become bidirectional and the application would consume events instead of bytes.

In a three layer model. 
 - App
 - PlatformClient
 - Core

The application shouldn't have to perform polling, and preferably wouldn't.

jazzz: My belief is that clients should be handling as much as possible. 

To make this function the client would need to become more tightly coupled with a single implemention however this is not necessarily bad. Specifically if the focus is on LogosDelivery. 

Specifically choosing concrete implementations as the client level makes the system easier, so i would support going that direction if desired

docs/adr/0001-client-event-system.md
osmaczko: > Does this ADR focus both scenarios or only one of the scenarios?

It focuses on the latter. The ADR is about how to surface events from `libchat`, through `client`, to the `app` or just from the `client` to the `app`.

> 1. the events that `libchat` gets notified when `logos-delivery` has state changes, for example, message been acked via SDS.

How external events, i.e. service events, are handled is not specified here. I see them being handled by the `client` through a channel receiver and then propagated to `libchat` through a function call if needed.

Take [this](https://github.com/logos-co/logos-chat-module/blob/9992d59a044d8e302112c2d321bb8fcb0a31dc2f/src/actions.rs#L98-L109) as an example:

```rust
    let conn_rx = match dm.on("connectionStateChanged") {
        Ok(rx) => Some(rx),
        Err(e) => {
            // Non-fatal: messaging still works; we just won't surface
            // delivery_state changes pushed by the node.
            eprintln!("chat_module_init: subscribe(connectionStateChanged) failed: {e}");
            None
        }
    };
```


Then, if relevant, this information can be propagated to `libchat`. Abstract example:
```rust
fn handle_connection_state(status: &str, detail: &str) {
    if status == "online" { libchat.notify_status(detail); }
}
```

docs/adr/0001-client-event-system.md
osmaczko: https://github.com/logos-messaging/libchat/pull/99#discussion_r3248933482

osmaczko:
@jazzz @kaichaosun I've addressed the comments (see last commit), please re-review.

jazzz (approved):

PR #100, opened 2026-05-19
Dependency cleanup

jazzz:
This PR standardizes cargo dependencies across the different crates. The lack of standardization was leading to merge conflicts and difficulty in tracing through the codebase. 

- Workspace crates which are used in more than one location are added as workspace dependencies 
- Dependencies are sorted, to minimize conflicts in future PRs
- sqlite, and components updated to use their package name rather than an alias 
- client crate renamed to `logos-chat` to clarify that it is the top level entry point for developers, and to reduce conflicts in developer applications

### DIFF
--- a/bin/chat-cli/src/app.rs
+++ b/bin/chat-cli/src/app.rs
@@ -5,7 +5,7 @@ use std::sync::mpsc;
 
 use anyhow::Result;
 use arboard::Clipboard;
-use client::{ConversationIdOwned, DeliveryService};
+use logos_chat::{ConversationIdOwned, DeliveryService};
 use serde::{Deserialize, Serialize};
 
 use crate::utils::now;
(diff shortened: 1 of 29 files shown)


bin/chat-cli/src/app.rs
```diff
@@ -42,7 +42,7 @@ pub struct AppState {
 }
 
 pub struct ChatApp<D: DeliveryService> {
-    pub client: client::ChatClient<D>,
+    pub client: logos_chat::ChatClient<D>,
```
kaichaosun: We can remove the crate name if there is no conlict for `ChatClient`.

kaichaosun (approved):

bin/chat-cli/src/app.rs
jazzz: Completely agree. I left it so the changeset was only replacements, and reorders. But I'll update it before I merge if there is no pushback

osmaczko (approved):

PR #102, opened 2026-05-19
Add crate logos-accounts

jazzz:
Accounts were introduced in a previous PR however they were local to the core/conversations. This PR moves Accounts to their own crate so they can be re-used in other services.

Changes:
- Adds IdentityProvider trait to `service_traits` so it is available for external implementors. 
- Adds logos-account crate, as it need to be accessible by applications to administer user accounts. This change while perhaps permature, establishes it as an external entity. 
- Updated IdentityProvider::friendly_name -> display_name as suggested by @osmaczko .

Future Work:
- Update core::context to accept an external IdentityProvider impl on construction. 
- Update InboxV2 to use this


jazzz:
This PR is dependent on #101  being merged first.

osmaczko (approved):

PR #103, opened 2026-05-20
Update InboxV2 to use IdentityProvider

jazzz:
This PR is the sibling PR to #102. It updates InboxV2 to use and abstract IdentityProvider, rather than requiring the internal LogosAccount implementation.

## Changes
- Removes the direct dependency on LogosAccount within InboxV2 and GroupV1 in preparation for its removal. 
- Remove MLSContext as suggested by @kaichaosun. Instead the functionality is flattened to mimic a OpenMLSProvider. This enables future implementation of MLS storage. 

## Deferred Work:
- The context interface is left unchanged. Once this and #102 land, the interface, examples and tests can be updated. This keeps the PR smaller.
- Removal of conversations::LogosAccount. The internal representation is still used. It can be removed once the Context constructors accept external IdentityProvider Impls.

## Notes
- This Depends on #101 so its currently marked as Draft

## Future PRs
- Remove generics at in GroupV1.
- Remove Runtime borrow checks in GroupV1

### DIFF
--- a/core/conversations/src/inbox_v2.rs
+++ b/core/conversations/src/inbox_v2.rs
@@ -1,10 +1,14 @@
+mod identity;
+mod mls_provider;
+
+pub use identity::MlsIdentityProvider;
+
 use std::cell::RefCell;
 use std::rc::Rc;
 
 use chat_proto::logoschat::envelope::EnvelopeV1;
 use openmls::prelude::tls_codec::Serialize;
 use openmls::prelude::*;
-use openmls_libcrux_crypto::Provider as LibcruxProvider;
 use prost::{Message, Oneof};
 use storage::ChatStore;
 use storage::ConversationKind;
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -10,6 +10,7 @@ members = [
     "core/double-ratchets",
     "core/integration_tests_core",
     "core/sqlite",
+    "core/sqlite",
```
kaichaosun: Duplicated with above one.

core/conversations/src/inbox_v2.rs
```diff
@@ -74,40 +34,55 @@ fn conversation_id_for(account_id: &AccountId) -> String {
...
     CS: ChatStore,
 {
     pub fn new(
-        account: LogosAccount,
+        account: IP,
         ds: Rc<RefCell<DS>>,
         reg_service: Rc<RefCell<RS>>,
         store: Rc<RefCell<CS>>,
     ) -> Self {
+        // Avoid referencing a temporary value by caching it.
```
kaichaosun: Is this comment still valid/

core/conversations/src/inbox_v2.rs
```diff
@@ -74,40 +34,55 @@ fn conversation_id_for(account_id: &AccountId) -> String {
...
 /// An PQ focused Conversation initializer.
 /// InboxV2 Incorporates an Account based identity system to support PQ based conversation protocols
 /// such as MLS.
-pub struct InboxV2<DS, RS, CS> {
+pub struct InboxV2<IP, DS, RS, CS>
+where
+    IP: IdentityProvider,
+{
     account_id: AccountId,
+    account: Rc<RefCell<MlsIdentityProvider<IP>>>,
```
kaichaosun, line 62: It should be identity, right?
Also I'm wondering if we can derive `account_id` from IdentityProvider?

core/conversations/src/inbox_v2/mls_provider.rs
```diff
@@ -0,0 +1,77 @@
...
+/// This is a Post-Quantum based MLS provider with in memory storage
+pub struct MlsEphemeralPqProvider {
+    crypto: LibcruxCryptoProvider,
+    storage: MemoryStorage,
+}
+
+impl MlsEphemeralPqProvider {
+    pub fn new() -> Result<Self, CryptoError> {
+        let crypto = LibcruxCryptoProvider::new()?;
+        let storage = MemoryStorage::default();
```
kaichaosun, line 24: Should we inject the storage, so it can be easier to be replaced later with actual store engine.

core/conversations/src/inbox_v2.rs
jazzz: Hmm great point. We need to lock down our language for the Accounts/Identity Model. Will handle offline

osmaczko (approved):

core/conversations/src/inbox_v2/mls_provider.rs
jazzz: This is only ever initialized in a single location - Its east to change the types. 

Skipping injection for now, as we can solve the problem later.

core/conversations/src/inbox_v2.rs
jazzz: Yes. but its obviously not clear enough. 

Its referring to that `account_id` being a class member so references can be returned. 

I'll move the comment and make it more clear

kaichaosun (approved):
Good to merge once clearness is finished.

PR #104, opened 2026-05-20
feat: introduce client event system

osmaczko:


core/conversations/src/service_traits.rs
```diff
@@ -1,18 +1,31 @@
...
+pub trait DeliveryService: Debug + Send + Sync {
     type Error: Display + Debug;
-    fn publish(&mut self, envelope: AddressedEnvelope) -> Result<(), Self::Error>;
-    fn subscribe(&mut self, delivery_address: &str) -> Result<(), Self::Error>;
+    fn publish(&self, envelope: AddressedEnvelope) -> Result<(), Self::Error>;
+    fn subscribe(&self, delivery_address: &str) -> Result<(), Self::Error>;
+
+    /// Return every inbound payload that has arrived since the last call.
+    /// Non-blocking; returns an empty vec when nothing is available.
+    fn pull(&self) -> Vec<Vec<u8>>;
```
jazzz, line 28: I see a few things happening here. 

1) I see the DS changing to use Interior mutability (Mutref -> ref)
2) The addition of a Pull function. 

It would take some work to convince me this is the preferred approach - specifically number 2. 

I would expect that traits reflect the functionality needed at a given level in the architecture. I would expect that the layer performing the the poll would add additional requirements if needed, and expose them to its providers, rather than adding it as a blanket requirement everywhere. 

"Do we really want ConversationTypes to call Poll?"

core/conversations/src/context.rs
```diff
@@ -9,23 +7,29 @@ use crate::{DeliveryService, RegistrationService};
...
+
 // This is the main entry point to the conversations api.
 // Ctx manages lifetimes of objects to process and generate payloads.
 pub struct Context<DS: DeliveryService, RS: RegistrationService, CS: ChatStore> {
-    identity: Rc<Identity>,
-    ds: Rc<RefCell<DS>>,
-    store: Rc<RefCell<CS>>,
+    identity: Arc<Identity>,
+    ds: Arc<DS>,
+    store: Arc<Mutex<CS>>,
```
jazzz, line 32: Making Context thread safe seems like its adding complexity to our pipeline - but perhaps there is a good reason for it. 

In general "Context" is getting too state aware, and I'd like to see ownership move up the the platform client level which could handle threading, io, etc. 

In the ServiceContext model these would all be passed in on invocation. 
So similar to this: [Example](https://github.com/logos-messaging/libchat/blob/main_mls/core/core_client/src/core_client.rs#L139-L147) however it doesn't own the ServiceContext - Platform Client does. This removes all runtime borrow checks, making the core synchronous and simple.

I think its fair that the Context/CoreClient is difficult to use, and the platform client provides the DX sugar, and owns the Services.

core/conversations/src/conversation.rs
```diff
@@ -43,11 +42,15 @@ pub trait Convo: Id + Debug {
...
-    fn send_content(&mut self, content: &[u8]) -> Result<(), ChatError>;
+    /// Invite new members and publish the resulting MLS commit + welcome
+    /// messages. Returns the observation events generated while publishing —
+    /// typically empty on success, `DeliveryFailed` if a publish errored.
+    fn add_member(&mut self, members: &[&AccountId]) -> Result<Vec<Event>, ChatError>;
+
+    /// Encrypt and publish `content`. This is the publish-side counterpart of
+    /// `Convo::send_message`. Returns the observation events generated while
+    /// publishing — typically empty on success.
+    fn send_content(&mut self, content: &[u8]) -> Result<Vec<Event>, ChatError>;
```
jazzz, line 53: Returning events makes sense to me. 

Theres an option to add a EventService, where clients can push events. However I like that this is a clean type driven approach.

core/conversations/src/event.rs
```diff
@@ -0,0 +1,56 @@
...
+        data: Vec<u8>,
+    },
+    #[non_exhaustive]
+    DeliveryFailed {
+        conversation_id: ConversationIdOwned,
+        /// `None` when the failure isn't tied to a specific outbound envelope.
+        envelope_id: Option<EnvelopeId>,
+        reason: FailureReason,
+    },
+}
```
jazzz, line 36: To maintain our deterministic parsing tree approach:

I'd love for our Event structure to capture in a self-describing way which events are destined for the Application. When I go to spec this, I want there to be no ambiguity whether an event should be passed to an App or consumed by an internal layer. 

If "All Events MUST be passed to application" that works.  However I want to avoid cases where a new Event is added in the future which creates ambiguity.

core/conversations/src/event.rs
```diff
@@ -0,0 +1,56 @@
...
+        conversation_id: ConversationIdOwned,
+        data: Vec<u8>,
+    },
+    #[non_exhaustive]
+    DeliveryFailed {
+        conversation_id: ConversationIdOwned,
+        /// `None` when the failure isn't tied to a specific outbound envelope.
+        envelope_id: Option<EnvelopeId>,
+        reason: FailureReason,
+    },
```
jazzz, line 35: We will likely need some clarity between how Events and Errors Differ assuming we still have both

jazzz (commented):
Only a quick glance as this is a draft.

Overall I can see this PR is making some much needed improvements. Specifically addressing ownership and simplifying execution pathways.


Few notes:
- This PR seems much bigger than it needs to be. Looks like we could just focus on event passing from the core, and leave DS updates until a later PR.
- I'd like to see the platform client be the object which is async/thread aware. Even if that means that Client uses async await via Tokio. 
- Instead of adding more functionality to traits consider defining another one which describes additional behavior. eg. `pub trait CanFetch:  fn fetch(.....` this makes managing traits easier.

core/conversations/src/event.rs
osmaczko: Good point, addressed in https://github.com/logos-messaging/libchat/pull/106 by two event types, one per layer.

osmaczko:
Thanks for review @jazzz 

> * This PR seems much bigger than it needs to be. Looks like we could just focus on event passing from the core, and leave DS updates until a later PR.

Agreed. Extracted this part to a separate PR: https://github.com/logos-messaging/libchat/pull/106

osmaczko:
Suppressed by #106 and #125

Issue #98, opened 2026-05-13
Group participants in a conversation can tell if a message is missing, and who sent it.


kaichaosun:
R1 in group chat [deliverables](https://roadmap.logos.co/messaging/roadmap/milestones/2025-extend-chat-sdk-with-group-conversations#deliverables). 

Delivery notification integration with Chat, relates to issue https://github.com/logos-messaging/libchat/issues/97.

Tasks included:
- SDS latest updates
- the events which is cared in group chat
- propose different solutions and comparison
- integrate with groupv1 chat

kaichaosun:
The proposal: 
https://www.notion.so/Reliable-Messaging-in-Libchat-3678f96fb65c80a5acd8ff8d10c7539a#3678f96fb65c809589a4ff2f3ca276c9

PR #105, opened 2026-05-21
feat: causal history notify missing messages

kaichaosun:
Related to issue: 
https://github.com/logos-messaging/libchat/issues/98

Proposed solution with option 2 in doc, but without much refining to the sds spec about causal history part:
https://www.notion.so/Reliable-messaging-in-libchat-3678f96fb65c80a5acd8ff8d10c7539a

### DIFF
--- a/core/conversations/src/causal_history.rs
+++ b/core/conversations/src/causal_history.rs
@@ -1,0 +1,259 @@
+//! Causal-history tracking for group conversations.
+//!
+//! Implements the *causal history* subset of the Scalable Data Sync (SDS)
+//! protocol. Every outbound message carries a Lamport timestamp and the IDs of the
+//! messages its sender had most recently seen. A receiver that finds a
+//! referenced ID it has never delivered knows a message is missing.
+//!
+//! Scope:
+//!  - assign a deterministic message ID + Lamport timestamp to outbound msgs
+//!  - attach a bounded causal-history frontier to each outbound message
+//!  - on receive, detect referenced-but-unseen message IDs (gaps)
+//!
+//! Out of scope here: bloom-filter acknowledgements,
+//! resend / outgoing buffer, incoming reorder buffer, Store-based recovery.
+//! This is detection only — an out-of-order message is still delivered to
+//! the application, but the gap it implies is reported.
+//!
+//! State is in-memory and session-scoped, matching the crate's current
+//! in-memory MLS state.
+
+use std::cell::RefCell;
+use std::collections::{HashMap, HashSet, VecDeque};
+use std::rc::Rc;
+
+use crate::proto::{Bytes, HistoryEntry, ReliablePayload};
+use crate::utils::{blake2b_hex, hash_size};
+
+/// Number of most-recently-seen message IDs attached to each outbound message.
+const CAUSAL_HISTORY_LEN: usize = 10;
+
+/// A message detected as missing: referenced by a delivered message's causal
+/// history but never seen locally.
+///
+/// This is the hook point for the future client event system (issue #97);
+/// until that lands, callers drain these via [`CausalHistoryStore::take_missing`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MissingMessage {
+    pub conversation_id: String,
+    pub message_id: String,
+}
+
+/// Per-conversation causal state.
+#[derive(Debug, Default)]
+struct ConvoState {
+    /// Lamport logical clock.
+    lamport_clock: i32,
+    /// Every message ID delivered locally (own sends + received).
+    seen: HashSet<String>,
+    /// Bounded frontier of recently-seen IDs (oldest first) attached to
+    /// outbound messages as causal history.
+    frontier: VecDeque<String>,
+    /// Missing IDs already reported, so a gap is surfaced exactly once.
+    reported_missing: HashSet<String>,
+}
+
+impl ConvoState {
+    fn record_seen(&mut self, id: String) {
+        if self.seen.insert(id.clone()) {
+            self.frontier.push_back(id);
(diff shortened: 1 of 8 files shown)


core/conversations/src/conversation/group_v1.rs
```diff
@@ -7,13 +7,16 @@ use std::rc::Rc;
 
 use blake2::{Blake2b, Digest, digest::consts::U6};
 use chat_proto::logoschat::encryption::{EncryptedPayload, Plaintext, encrypted_payload};
+use chat_proto::logoschat::reliability::ReliablePayload;
 use crypto::Ed25519VerifyingKey;
 use openmls::prelude::tls_codec::Deserialize;
 use openmls::prelude::*;
 use openmls_libcrux_crypto::Provider as LibcruxProvider;
 use openmls_traits::signatures::Signer as OpenMlsSigner;
+use prost::Message as _;
```
jazzz, line 16: Cute. I like it

core/conversations/src/causal_history.rs
```diff
@@ -0,0 +1,259 @@
...
+/// A message detected as missing: referenced by a delivered message's causal
+/// history but never seen locally.
+///
+/// This is the hook point for the future client event system (issue #97);
+/// until that lands, callers drain these via [`CausalHistoryStore::take_missing`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MissingMessage {
+    pub conversation_id: String,
+    pub message_id: String,
+}
```
jazzz, line 40: Love this - Great work thinking ahead and avoiding wasted effort

core/conversations/src/causal_history.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// Detected gaps, drained by the client (future #97 event bus).
+    missing: Vec<MissingMessage>,
+}
+
+/// Session-scoped causal-history store shared by every `GroupV1Convo`
+/// instance.
+///
+/// Convos are rebuilt from storage on every inbound message, so this state
+/// cannot live on the convo struct — it is shared through `InboxV2`, the
+/// same way the MLS provider is.
```
jazzz, line 79: True - I think we'll need to look at that soon. Once DeMLS lands. 

Having CasualHistoryStore external service to the Convo is not ideal, but I agree its the best path for now.

core/conversations/src/causal_history.rs
```diff
@@ -0,0 +1,259 @@
...
+        b"sds-msg-id|".as_slice(),
+        channel_id.as_bytes(),
+        b"|".as_slice(),
+        sender.as_bytes(),
+        b"|".as_slice(),
+        lamport_be.as_slice(),
+        b"|".as_slice(),
+        content,
+    ])
+}
```
jazzz, line 185: I think this is smart approach. Specifically extracting the 4 tuple. Making the properties of this identifier independent and standalone will serve us well.

core/conversations/src/utils.rs
```diff
@@ -40,6 +40,8 @@ pub mod hash_size {
         AccountId => U8,
         /// Conversation ID hash length
         ConvoId  => U6,
+        /// Causal history message ID hash length (256-bit, collision-resistant)
+        MessageId => U32,
```
jazzz, line 44: 🙏  - I love being able to see all the the different uses of Identifiers in our system.

core/conversations/src/causal_history.rs
```diff
@@ -0,0 +1,259 @@
...
+/// Deterministic, collision-resistant message ID.
+///
+/// A single sender increments its Lamport clock on every send, so
+/// `(sender, lamport)` is unique per message; `channel_id` and `content` are
+/// folded in as well. Receivers store the field verbatim, so cross-peer
+/// agreement does not depend on re-derivation.
+fn derive_message_id(channel_id: &str, sender: &str, lamport: i32, content: &[u8]) -> String {
+    let lamport_be = lamport.to_be_bytes();
+    blake2b_hex::<hash_size::MessageId>(&[
+        b"sds-msg-id|".as_slice(),
```
jazzz: > "sds-msg-id" 

1) Domain separator is not needed in this context, however hard to argue against using one. Having all hashes tied to their context is a development practice I can get behind. 👍 

2) [Dust] I'd decouple this from SDS It appears to have more value and is something I'd want to reuse else where. Perhaps "deterministic_frame_id"?

core/conversations/src/conversation/group_v1.rs
```diff
@@ -264,9 +281,14 @@ where
...
+
+        let sender_id = self.account_id.as_str();
+        let reliable = self.causal.on_send(&self.convo_id, sender_id, content);
+        let wire = reliable.encode_to_vec();
+
         let mls_message_out = self
             .mls_group
-            .create_message(provider, ctx_ref.ident(), content)
+            .create_message(provider, ctx_ref.ident(), &wire)
             .unwrap();
```
jazzz, line 292: [Note] In the future we'll want this to apply to all frames, not just Content. That infra doesn't exist for now, but highlighting for the future.

jazzz (approved):
There is a lot to LOVE in this PR. 

- Changes are clean. 
- Well thought out design decisions around Identifiers.
- Fast & Efficient - This makes the minimal changes  and allows for future follow ups once we know more. 


This is great work.

jazzz (approved):
There is a lot to LOVE in this PR. 

- Changes are clean. 
- Well thought out design decisions around Identifiers.
- Fast & Efficient - This makes the minimal changes  and allows for future follow ups once we know more. 


This is great work.

core/conversations/src/conversation/group_v1.rs
kaichaosun: Noted, thanks for pointing out earlier.

PR #106, opened 2026-05-21
feat: introduce client event system

osmaczko:
- Core processing yields a `PayloadOutcome` enum — `Empty`, `Convo`, or
  `Inbox`. `ConvoOutcome` carries a conversation id and an optional
  decrypted `Content`; `InboxOutcome` adds a `NewConversation`
  (id + `ConversationClass`) for a peer-initiated conversation.
- Client translates `PayloadOutcome` into app-facing `Vec<Event>`
  (`ConversationStarted`, `MessageReceived`) at the boundary, so the
  application loop sees discrete events rather than core types.
- MLS group welcomes produce a `ConversationStarted` event with no
  initial content, fixing the silent-group-join case where the inbox
  layer dropped the observation.
- C FFI exposes an `EventList` opaque type with indexed accessors and
  an `Invalid` sentinel for out-of-bounds / non-applicable reads.
- Symmetric `Inbox` / `InboxV2` handlers: both return
  `Result<InboxOutcome, _>` and own the persistence + ephemeral-key
  cleanup for the conversations they create.
- Updated and simplified `docs/adr/0001-client-event-system.md`.


core/conversations/src/conversation/group_v1.rs
```diff
@@ -316,20 +316,19 @@ where
             .map_err(ChatError::generic)?;
 
         match processed.into_content() {
-            ProcessedMessageContent::ApplicationMessage(msg) => Ok(Some(ContentData {
-                conversation_id: hex::encode(self.mls_group.group_id().as_slice()),
-                data: msg.into_bytes(),
-                is_new_convo: false,
-            })),
+            ProcessedMessageContent::ApplicationMessage(msg) => Ok(vec![Event::MessageReceived {
+                convo_id: Arc::from(self.id()),
```
jazzz: ~[Sand]~ Rather than introducing reference counting, the event can own the convo_id.

core/conversations/src/conversation/privatev1.rs
```diff
@@ -253,13 +242,13 @@ impl<S: ConversationStore + RatchetStore> Convo for PrivateV1Convo<S> {
...
-        };
-
-        Ok(output)
+        Ok(match frame_type {
+            FrameType::Content(bytes) => vec![Event::MessageReceived {
+                convo_id: Arc::from(self.id()),
+                content: bytes.into(),
+            }],
+            FrameType::Placeholder(_) => vec![],
+        })
```
jazzz: [Sand] In general we'd like to see symmetric parsing patterns for easier readability.

The previous recipe remains readable and symmetric as more frames are added. Each type gets its own function, match statement maps FrameTypes to handler functions:
```
  let output = match frame_type {
            FrameType::Content(bytes) => self.handle_content(bytes.into()),
        //  FrameType::Reset(bytes) => self.handle_reset(bytes.into()),
            FrameType::Placeholder(_) => None,
        };
```

The current approach removes the dedicated function, which causes more nesting.
```
Ok(match frame_type {
            FrameType::Content(bytes) => vec![Event::MessageReceived {
                convo_id: Arc::from(self.id()),
                content: bytes.into(),
            }],
            FrameType::Placeholder(_) => vec![],
        })
```

Prefer flatter shorter functions, and symmetry where possible

core/conversations/src/conversation/group_v1.rs
jazzz: Ah I see this comes from the event using ` type ConversationIdOwned = Arc<str>;` This makes sense for now.

core/conversations/src/inbox/handler.rs
```diff
@@ -152,16 +155,20 @@ impl<S: EphemeralKeyStore> Inbox<S> {
...
+                    return Err(ChatError::Protocol(
+                        "expected initial message in invite".into(),
+                    ));
+                }
+
+                let started = Event::ConversationStarted {
+                    convo_id: Arc::from(convo.id()),
+                    kind: ConversationKind::PrivateV1,
                 };
+                let events = std::iter::once(started).chain(inner_events).collect();
```
jazzz: [?] I see that events is being built from [Started, .. inner_events]. Which seems like a deliberate choice, vs extending inner_events.

It raises the question:

Does Event order matter? Are there rules that need to be followed to make events function?

core/conversations/src/context.rs
```diff
@@ -313,7 +305,7 @@ where
             ConversationKind::GroupV1 => Ok(Box::new(
                 self.pq_inbox.load_mls_convo(record.local_convo_id)?,
             )),
-            ConversationKind::Unknown(_) => Err(ChatError::BadBundleValue(format!(
+            _ => Err(ChatError::BadBundleValue(format!(
```
jazzz: [?] This is a semantic change to how ConversationKinds are handled. This changes from a compile time error for unhandled variants, to a runtime thrown error. 

Is there a driver for this change?

jazzz (commented):
This looks like a good change to me. 

If @kaichaosun  is onboard then so am I

core/conversations/src/conversation/group_v1.rs
```diff
@@ -281,10 +284,7 @@ where
         Ok(vec![a])
     }
 
-    fn handle_frame(
-        &mut self,
-        encoded_payload: EncryptedPayload,
-    ) -> Result<Option<ContentData>, ChatError> {
+    fn handle_frame(&mut self, encoded_payload: EncryptedPayload) -> Result<Vec<Event>, ChatError> {
```
kaichaosun: Return `Vec<Event>` looks overkill for a specific frame handling.
IMO the conversation should take minimal knowledge about outside of the core.
But Event implicitly hints there is an event process system somewhere, and such procedure is outside of core. And a vector of events brings even more unclearness.

It's ok if it's one layer, but we intentionally design different layers, app -> client -> core. It's better to have each layer's ontology / terms.

If the intention is to build an event driven system as an overall architecture, we need to do research if it's possible in core's synchronous restriction, and there may be requirements for framework for such an architecture.

core/conversations/src/context.rs
```diff
@@ -337,7 +329,7 @@ where
...
```
kaichaosun: We should exhaustic the enum when possible.

core/conversations/src/event.rs
```diff
@@ -0,0 +1,28 @@
...
+//! on existing variants are additive.
+
+use storage::ConversationKind;
+
+use crate::context::ConversationIdOwned;
+
+/// A discrete observation produced by processing an inbound payload.
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+pub enum Event {
```
kaichaosun: Refer to this comment: https://github.com/logos-messaging/libchat/pull/106#discussion_r3285474869

docs/adr/0001-client-event-system.md
```diff
@@ -209,65 +225,101 @@ pub fn tick(&mut self) -> Result<()> {
...
 
-| Variant | Emitted from |
+| `logos_chat::Event` variant | Source |
 |---|---|
-| `ConversationStarted` (responder side) | `core/conversations/src/inbox/handler.rs:155-162` (replaces `is_new_convo: true`) |
-| `MessageReceived` | `core/conversations/src/conversation/privatev1.rs:184-191` (replaces `is_new_convo: false`) |
-| `DeliveryReceipt` | `Context::handle_payload` when decoding a `PrivateV1Frame::Receipt` (future protocol work) |
-| `DeliveryFailed { Transport }` | Core method that invoked `delivery.publish` and observed a synchronous error |
-| `DeliveryFailed { Timeout }` | Client's background retry thread |
+| `ConversationStarted` | Wrapped from `libchat::Event::ConversationStarted` in `ChatClient::receive` |
```
kaichaosun: Keep the doc updated along with the code can be hard. It's better to just give the general context and final decision about the design, not much detail about code.

crates/client/src/event.rs
```diff
@@ -0,0 +1,26 @@
...
+#[derive(Debug, Clone)]
+pub enum Event {
+    /// A new conversation has appeared.
+    #[non_exhaustive]
+    ConversationStarted {
+        convo_id: ConversationIdOwned,
+        kind: ConversationKind,
+    },
+    /// User content arrived on an existing conversation.
+    #[non_exhaustive]
```
kaichaosun: Wondering why non-exhaustive on variant is needed? The struct here looks business heavy, add a new field likely change the way to process the struct.

crates/client/src/client.rs
```diff
@@ -74,13 +75,25 @@ impl<D: DeliveryService + 'static> ChatClient<D> {
...
+    /// may be empty for protocol-only frames.
+    pub fn receive(&mut self, payload: &[u8]) -> Result<Vec<Event>, ClientError<D::Error>> {
+        // Explicit per-variant match rather than `From`: core and client
+        // event shapes are allowed to diverge.
+        Ok(self
+            .ctx
+            .handle_payload(payload)?
+            .into_iter()
+            .filter_map(|e| match e {
+                CoreEvent::ConversationStarted { convo_id, kind, .. } => {
```
kaichaosun: Such type conversion looks interleaved between the two layers, and hurts split of concern principle.

core/conversations/src/conversation/group_v1.rs
jazzz: > Return Vec<Event> looks overkill for a specific frame handling.
IMO the conversation should take minimal knowledge about outside of the core.

I'm not sure I am following this. 

I agree that Conversations should know as little as possible about anything outside of it. It defines what it needs, it defines what it provides.

jazzz: A Vec of Events aka "ReturnObjects" - seems logical to me. But perhaps I am missing something. 

The meaningful change is that ContentData has been expanded to an enum that supports other things. I'd have to verify however its entirely possible that an inbound frame generates content, and notifies the client that a message in the causal change is missing. How would you suggest to capture that information?

core/conversations/src/context.rs
jazzz: Interested if @osmaczko  has any pushback here. [I'd agree](https://github.com/logos-messaging/libchat/pull/106#discussion_r3282546056), especially within the core. We want errors at compile time if possible, not runtime.

docs/adr/0001-client-event-system.md
```diff
@@ -209,65 +225,101 @@ pub fn tick(&mut self) -> Result<()> {
...
 
 2. **Consumer pattern: Option B — channel handed to the caller.** Different consumer archetypes could favour different shapes — a polling UI loop suits Option A; a low-latency push-driven consumer (toast notifications, daemons) suits Option C — but Option B is preferred: it is the most Rust-idiomatic of the three, has few drawbacks compared to A or C, and offers the smoothest path to async (point 1).
 
+3. **Two event types, mapping at the client boundary.** `libchat::Event` carries observations the core produced; `logos_chat::Event` is the app-facing type. The client maps between them in an explicit per-variant `match`, with the freedom to wrap, transform, aggregate, or drop. "App-bound or not" is a property of the type, not of a per-variant predicate, so there is no ambiguity about whether a given variant is meant for the application. The two enums can diverge as needed: client-only events like `DeliveryFailed { Timeout }` never leak into the core; core-only observations never leak to the app. No blanket `From`/`TryFrom` impl, because a 1-for-1 correspondence is not what we want to commit to.
+
```
jazzz: I like this. It provides a clean break between the core and the client. 

For being clear about alignment:
- We'll specify the events that come from the core. They are a interop point.
- Client event compartmentalization would be a implementation suggestion. 
   - Clients can be opinionated in what data they provide to applications. 
   - We should definitely implement it that way in the reference client, but from a protocol focus its definitely out of scope.

I only reiterate this to highlight that we are working on two separate products at the same time, and each has its own requirements/properties.

jazzz: [Dust] namespacing works, but its harder to reason about in documentation. I'd suggest considering explicit names. ([Core|Convo]Event, [Chat|App]Event). Different layers, different names, precision through embedded context.

crates/client-ffi/src/api.rs
```diff
@@ -26,6 +27,43 @@ pub enum ErrorCode {
...
+#[repr(i32)]
+#[derive(Clone, Copy)]
+pub enum FfiConversationKind {
+    /// Sentinel for accessor calls that don't apply to the queried row
+    /// (out-of-bounds, or a non-`ConversationStarted` event).
+    Invalid = -1,
+    Unknown = 0,
+    PrivateV1 = 1,
+    GroupV1 = 2,
+}
```
jazzz, line 55: [Pebble] This seems like the wrong approach. Though you may have your reasoning for taking this path for now. Just highlighting the anti-pattern. 

From a conceptual level. This is abstraction leakage from the core.

Clients: are purposely not aware of the specific ConversationType used: https://github.com/logos-messaging/libchat/blob/cbe9e532b25ad06336476de97eb2f7f2508d7ff2/crates/client/src/client.rs#L52-L61.

The purpose behind this is; it does not matter. Exposing these details makes it an Application level concern. Which means that future additions of conversations could break clients and applications. 

The 'class' of conversation matters (Private, Group, Broadcast) as they have expose different functionality. But all instances of a class should remain functionally the same. 

I there still work to have the required stability, however avoiding exposing data that does not have a direct functional impact will keep the code base leaner. 

The goal is that GroupVX can eventually be added with only a PR to core. 
- Add ConversationType
- Ensure its storage requirements are met.
- Update context::create_group_convo to instantiate the new conversation type.

crates/client-ffi/src/api.rs
```diff
@@ -44,14 +82,75 @@ pub struct CreateConvoResult {
...
+enum EventRow {
+    ConversationStarted {
+        convo_id: String,
+        kind: FfiConversationKind,
+    },
+    MessageReceived {
+        convo_id: String,
+        content: Vec<u8>,
+    },
+}
```
jazzz, line 103: [Wish] Having a whole set of types specifically for FFI is unreasonably annoying me. I fully understand that the code surface is small, and its maintainability is reasonable. I get that this is the good approach - I have no other better solutions.

However the concept of writing specific rust code, to convert a rust type to another rust type, to then auto extract to a C representation is just infuriating. 

Consider this an open invite to lurkers - a better solution here would be much welcome.

jazzz: @osmaczko  does this get any better with: https://github.com/logos-co/logos-chat-module/pull/25

or should I create an issue and encourage contributions from the community?

crates/client-ffi/src/api.rs
```diff
@@ -26,6 +27,43 @@ pub enum ErrorCode {
...
+#[derive_ReprC]
+#[repr(i32)]
+#[derive(Clone, Copy)]
+pub enum EventKind {
+    /// Sentinel returned by `event_list_kind_at` for out-of-bounds indices.
+    /// Never the kind of a real event row.
+    Invalid = -1,
+    ConversationStarted = 0,
+    MessageReceived = 1,
+}
```
jazzz, line 44: [Dust] perfer const integers rather than an enum.

This seems excessive. There is no match on EventKind, its purely used to pass the the type to something that can be converted to a i32 for FFI. 

If the desired end state representation is a i32 just represent it as such and avoid the conversion functions and overhead.

The transitive mapping of. EventRow -> EventKind -> i32 seems like it doesn't add value, yet adds complexity.

jazzz (approved):
I think there is some clean up to address, but once you are happy with this I'm happy with this.

Some general feedback for the future #RelentlessImprovment

- Having PRs span multiple crates is hard to follow and and takes significantly more time to review. e.g. This PR Updates the Core, Client, FFI and test application.
   - Specifically the FFI updates are very easy for me to gloss over. I'm definitely not giving them the attention they deserve. 
   - If we can find a workflow that breaks these up that would be a huge improvement to the team.
   - Crawl, walk, run -> Core, Client, Rest
  
- Were starting a push for V0.2 so not an issue this month, but I'm starting to notice our functions are becoming deeper, busier, longer.
   - I'm not suggesting we strictly follow the '5 line function principal', but a soft reminder that readability is important.
   - Were looking for symmetry not necessarily short.
   - Good [steps](https://github.com/logos-messaging/libchat/pull/106/changes#diff-3dca3389b506129fd7ac0fb833b7742b4c81be1333f843d63843e684cc81013cR147-R149) in the right direction: 👍

core/conversations/src/conversation/group_v1.rs
kaichaosun: > however its entirely possible that an inbound frame generates content, and notifies the client that a message in the causal change is missing. How would you suggest to capture that information?

It seems we can return a struct to capture the new information like missing messages. If the field doesn't make sense in `privatev1`, I'm concerned that `handle_frame` interface may not make the right abstraction.

Another option is exposing new api for new information, but it may introduce unnecessary state management.

I see that introducing `Vec<Event>` is trying to make code more generic, but with a noticeable tradeoff, libchat-core can be hard to reason about by such tradeoff, especially with more custom handling logic for different convos.

core/conversations/src/context.rs
osmaczko: Overlooked Claude work, reverted, thanks for sharp eyes

docs/adr/0001-client-event-system.md
osmaczko: You are right, I've simplified the ADR significantly.

crates/client/src/event.rs
osmaczko: Good point, I'll keep only enum itself non-exhaustive, so we can add events without breaking applications.

crates/client/src/client.rs
osmaczko: This stands by design, see point 3 https://github.com/logos-messaging/libchat/blob/0983cc3205ae7d1c353951d056dca8941ef85807/docs/adr/0001-client-event-system.md#decisions.

This is the place, where client translates conversation observations to the application. See diagram: https://github.com/logos-messaging/libchat/blob/0983cc3205ae7d1c353951d056dca8941ef85807/docs/adr/0001-client-event-system.md#sequence.

I am open for suggestions.

core/conversations/src/inbox/handler.rs
osmaczko: No longer relevant, events are only on the client level and are aimed to be consumed in order they are produced.

core/conversations/src/conversation/group_v1.rs
osmaczko: That’s a good point, and it led me to re-architect the conversations part. Conversations are no longer aware of events, and they now use a structural result type.

core/conversations/src/context.rs
osmaczko: https://github.com/logos-messaging/libchat/pull/106#discussion_r3299127751

docs/adr/0001-client-event-system.md
osmaczko: Not relevant anymore, Conversations are no longer aware of events, and they now use a structural result type, as per updated ADR.

crates/client-ffi/src/api.rs
osmaczko: Good point! I've introduced [ConversationClass](https://github.com/logos-messaging/libchat/blob/addcde03eb8a3985c881f9e26a7d386b7f45a462/crates/client/src/client.rs#L121-L133) as a hint for applications.

crates/client-ffi/src/api.rs
osmaczko: The whole `client-ffi` is going to be removed once `logos-chat-module` is switched to logos core rust-sdk.

> @osmaczko does this get any better with: https://github.com/logos-co/logos-chat-module/pull/25

Partially. All information that `logos-chat-ui` needs will have to be exposed through C, which will probably be a subset of what is currently defined in `client-ffi`.

crates/client-ffi/src/api.rs
osmaczko: The transitive mapping is removed; keeping the enum because safer-ffi can't export pub const to the C header. I'll keep this hint in mind exposing C API from logos-module.

osmaczko:
@jazzz @kaichaosun thanks for review. I’ve reiterated over the solution to address your feedback. I’ve also simplified the ADR a lot, as it contained too many implementation details. Please get familiar with it first when attempting a re-review.

kaichaosun (approved):
Well done on the refactor 👍 .

I have also tried to build on your branch and see what other option exists, after some playaround, I got this [commit](https://github.com/logos-messaging/libchat/commit/572c5c9cd23b58a75669ae896974b50d73cd2a0e) in branch `feat/ces-v2-kc` with extra event `MessageMissing`.
It loses some generality with concrete types, but it offers more clarity. Feel free to take a look and see whether it makes sense in your view.

The underlying idea is that, before we have a mature protocol design, the code should not be overly abstracted.

Approved this PR since it's already working, further refactor can be handled later.

@osmaczko @jazzz

core/conversations/src/conversation/group_v1.rs
```diff
@@ -329,35 +331,35 @@ where
 
         if protocol_message.epoch() < self.mls_group.epoch() {
             // TODO: (P1) Add logging for messages arriving from past epoch.
-            return Ok(None);
+            return Ok(FrameOutcome::default());
```
jazzz: [Sand] Default should be reserved for opinionated initializations. Consider `::empty()` or `::new()`

core/conversations/src/conversation/group_v1.rs
```diff
@@ -329,35 +331,35 @@ where
...
             ProcessedMessageContent::ApplicationMessage(msg) => {
                 let reliable = ReliablePayload::decode(msg.into_bytes().as_slice())?;
                 self.causal.on_receive(&self.convo_id, &reliable);
-                Ok(Some(ContentData {
-                    conversation_id: hex::encode(self.mls_group.group_id().as_slice()),
-                    data: reliable.content.to_vec(),
-                    is_new_convo: false,
-                }))
+                vec![Message {
+                    convo_id: Arc::from(self.id()),
```
jazzz: [Sand] Use of Arc implies shared ownership. 

I think you want single ownership here.

core/conversations/src/inbox/handler.rs
```diff
@@ -119,14 +121,17 @@ impl<S: EphemeralKeyStore> Inbox<S> {
...
+    /// conversation and consumes the ephemeral key. Returns the
+    /// [`InboundResult`] describing what was observed — for a successful
+    /// invite, a `new_conversation` and one initial `message`.
     pub fn handle_frame<PS: ConversationStore + RatchetStore>(
         &self,
         enc_payload: EncryptedPayload,
         public_key_hex: &str,
         private_store: Rc<RefCell<PS>>,
-    ) -> Result<(Conversation<PS>, Option<ContentData>), ChatError> {
+    ) -> Result<InboundResult, ChatError> {
```
jazzz: [~~Pebble~~ Sand] Avoid mixing Inbox and ConverstationType processing. 

The two paths are different and should not result in the same result type. 

Inboxes are the only protocol objects that can create a ConversationType. Consider using a dedicated type for Inboxes or ignoring for now.

core/conversations/src/context.rs
```diff
@@ -237,53 +238,39 @@ where
...
-    ) -> Result<Option<ContentData>, ChatError> {
+    ) -> Result<InboundResult, ChatError> {
         let enc_payload = EncryptedPayload::decode(enc_payload_bytes)?;
         let mut convo = self.load_convo(convo_id)?;
-        convo.handle_frame(enc_payload)
+        let frame = convo.handle_frame(enc_payload)?;
+        Ok(InboundResult {
+            new_conversation: None,
+            frame,
+        })
```
jazzz: [Sand] Raw construction will be hard to manage. In this case the frame is ambivalent to the `new_conversation` field. 

Consider builder pattern or dedicated constructors.

core/conversations/src/inbound.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub convo_id: ConversationIdOwned,
+    pub kind: ConversationKind,
+}
+
+/// User content decrypted from an inbound payload.
+#[derive(Debug, Clone)]
+pub struct Message {
+    pub convo_id: ConversationIdOwned,
+    pub content: Vec<u8>,
+}
```
jazzz: [Pebble] Message is a banned identifier as far as I'm concerned - It does not reveal anything about the object and only leads to assumptions/confusion.

The term outlined in [Chatdefs](https://github.com/logos-messaging/specs/blob/master/informational/chatdefs.md) is Content

core/conversations/src/inbound.rs
```diff
@@ -0,0 +1,57 @@
...
+//!
+//! [`InboundResult`] composes two layers:
+//! - [`FrameOutcome`] captures what processing one frame within a conversation
+//!   produces: today, decrypted messages. As protocol features land, new
+//!   per-conversation observations (e.g. group membership changes) become
+//!   additive fields on `FrameOutcome`.
+//! - [`InboundResult`] wraps a `FrameOutcome` and adds the payload-level
+//!   observations a single frame cannot produce — today, the appearance of
+//!   a new conversation from the peer side.
+
```
jazzz: [Sand] `types.rs` is currently the home of all payload types that leave the core. 

consider merging to keep consistency. 

[Sand] "Inbound" as a module name makes me think processing, not the "outbound types" to the client.

core/conversations/src/inbound.rs
```diff
@@ -0,0 +1,57 @@
...
+use storage::ConversationKind;
+
+use crate::context::ConversationIdOwned;
+
+/// Observations a conversation produces from processing one frame.
+#[derive(Debug, Clone, Default)]
+pub struct FrameOutcome {
+    /// User content decrypted from this frame, in protocol order.
```
jazzz: [?] For my own reference what do we mean by "Protocol Order"?

core/conversations/src/inbound.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub messages: Vec<Message>,
+}
+
+impl FrameOutcome {
+    pub fn is_empty(&self) -> bool {
```
jazzz: [Dust] Empty implies a container. However FrameOutcome appears to be an Object representing the various observations. 

Either implement `iter` for consistency with other collections, or define what it means for an Outcome to be empty. 

If this object is not intended to grow, and purely represents a Vec of Content, then consider a typedef or a tuple struct which implements Deref

core/conversations/src/inbound.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+/// Everything one inbound payload produced.
+#[derive(Debug, Clone, Default)]
+pub struct InboundResult {
+    /// A new conversation appeared from this payload, if any.
+    pub new_conversation: Option<NewConversation>,
+    /// Observations from the frame inside this payload.
+    pub frame: FrameOutcome,
```
jazzz: [Pebble] the name 'frame' is misleading. 

A FrameOutcome does not contain frames, it contains Content. Frames SHOULD not bleed outside of the ConversationType abstraction. 

https://github.com/logos-messaging/specs/blob/master/standards/application/privatev1.md#architecture

core/conversations/src/inbound.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.new_conversation.is_none() && self.frame.is_empty()
+    }
+}
+
+/// A conversation newly observed from the peer side.
+#[derive(Debug, Clone)]
+pub struct NewConversation {
+    pub convo_id: ConversationIdOwned,
+    pub kind: ConversationKind,
```
jazzz: [Sand] This is a storage level concept which is now bleeding out into the client. 

From a lean design POV; Is there a direct need for clients to have access to ConversationKinds or any other meta data about the Conversation without polling for it?

docs/adr/0001-client-event-system.md
```diff
@@ -5,272 +5,55 @@
...
 
-Events are the uniform observation channel: they carry both observations the call itself caused (e.g. a sync `DeliveryFailed { Transport }` from `send_content`) and observations from background work (e.g. `DeliveryFailed { Timeout }` from the retry thread). The only thing kept outside `Vec<Event>` is an obvious primary result the caller will use immediately — returned directly for ergonomics. This is why the initiator side does not emit `ConversationStarted`: `create_private_convo` returns the new `ConversationIdOwned` directly as part of its return value.
+3. **Two enums, mapping at the client boundary.** `InboundResult` is the structural sum of observations from one payload; `Event` is a discrete app-facing notification. The two enums are allowed to diverge: a protocol-internal observation the app does not need lives only on `FrameOutcome`; a client-only event like `DeliveryFailed { Timeout }` lives only on `Event`. Translation is an explicit per-variant `match` inside the client — not a blanket `From` impl — to preserve that divergence as both sides grow.
```
jazzz: I like this a lot

docs/adr/0001-client-event-system.md
```diff
@@ -5,272 +5,55 @@
...
-    PeerRejected, // peer signalled rejection (future protocol work)
-    Timeout,      // no receipt within the retry window
-}
-```
-
-`#[non_exhaustive]` on the enum permits new variants; on each struct variant it permits new fields. Both are additive minor-release changes. Future variants (`ConversationRekeyed`, `ParticipantJoined`, `PresenceChanged`, transport health, key-rotation reminders, …) follow this rule.
+## Decisions
 
-Mapping of variants to emit sites:
+1. **Core returns `InboundResult`, a structural result type.** One field per kind of observation a payload can produce: an optional new conversation, plus a `FrameOutcome` carrying everything a per-conversation frame processor yields. The structural shape encodes causality (a new conversation is logically prior to anything that happens inside it), so a wrong ordering of observations cannot be represented in the type. `FrameOutcome` exists as a separate type because `Convo::handle_frame` cannot create a conversation; embedding it inside `InboundResult` keeps each return type producing only what its source can populate.
```
jazzz: > `FrameOutcome` exists as a separate type because `Convo::handle_frame` cannot create a conversation;

This makes sense. You want to be clear about what what are the valid return types for each Conceptual object.

docs/adr/0001-client-event-system.md
jazzz: Except I'm realizing that InboundResult was not implemented as an enum, but as a struct.

crates/client/src/client.rs
```diff
@@ -94,3 +93,41 @@ impl<D: DeliveryService + 'static> ChatClient<D> {
...
+/// deliberate mapping decision rather than silently misclassifying it.
+/// `Unknown(_)` yields `None`: the client does not surface conversations
+/// whose protocol kind cannot be safely classified for the application.
+fn class_from_kind(kind: &ConversationKind) -> Option<ConversationClass> {
+    match kind {
+        ConversationKind::PrivateV1 => Some(ConversationClass::Private),
+        ConversationKind::GroupV1 => Some(ConversationClass::Group),
+        ConversationKind::Unknown(_) => None,
+    }
+}
```
jazzz: [~~Pebble~~ Sand] This requires the client be updated when new conversations are being added. The aim should be to minimize this to a single location. Consider mapping in core. Core knows ConversationTypes, Client knows ConversationClass

crates/client/src/event.rs
```diff
@@ -0,0 +1,37 @@
...
+///
+/// Decoupled from the core's protocol-versioned kinds: future versions of
+/// an existing class (e.g. a `PrivateV2`) map to the same variant here.
+/// New variants are reserved for fundamentally different conversation
+/// shapes and are intentionally breaking when added.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ConversationClass {
+    Private,
+    Group,
+}
```
jazzz: Love this!

[Dust] Consider moving inside of the core boundary.

core/conversations/src/inbound.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/// Everything one inbound payload produced.
+#[derive(Debug, Clone, Default)]
+pub struct InboundResult {
+    /// A new conversation appeared from this payload, if any.
+    pub new_conversation: Option<NewConversation>,
+    /// Observations from the frame inside this payload.
+    pub frame: FrameOutcome,
+}
```
jazzz: [?] Comparing the previous and current approach in a flattened state I have some questions.

**ContentData**
```
Struct {
    convo_id
    Vec<u8>
    bool
}
```

**InboundResult**
```
Struct {
    Option<(convo_id)>
    Vec<convo_id, Vec<u8>>
}
```

Notes:
- `bool -->> Option< _ >` :: I think this makes sense
   - Option is clearer with the associated convo_id
   - However this isn't meaningfully different than before
- convo_id + Bytes  -->> Vec< convo_id , Bytes > :: This I'm not so sure about
   - Cardinality : Previously each inbound payload could generate at most one message. I'm not opposed to allowing multiple as a future proofing approach, but is there a usecase you can think of where this is needed?
   - ConvoMultiplexing : Previously each payload was associated to a single conversation. This format allows for FrameOutput to contain different convo_ids. Can you explain why that is needed? It appears to be a significant change to the parsing rules.

jazzz: I'm having a hard time seeing how InboundResult is a notable improvement over 


**ContentData++**
```
pub struct ProcessingOutcome {  // Change name
     convo_id: String,          // Keeps processing simple as conversation_id is enforced by type.
     content: Vec< Vec<u8> >,   // Allow multiple messages
     is_new_conversation: bool  // Option flattens to bool, as the identifier is already being used
}
```

Am I missing something important?

jazzz: If the blocker is restricting what Conversations can do compared to Inboxes and other objects.

[Sand] Consider using a dedicated enum for each pathway, and then implementing `From<_>` to define a mapping.

jazzz (commented):
I'm liking the clean up, and the iteration. 

This latest change has positive additions:
- Specific types for Specific paths.
- Client side events look great to me. 

I have questions about the backend that I'd like some answers to. 
There are also some Pebbles I'd like to understand the plan before an explicit approval.

With an action plan for future work I don't see a blocker here - specifically as the client side looks solid.

core/conversations/src/outcomes.rs
```diff
@@ -0,0 +1,82 @@
...
+        Self::Inbox(i)
+    }
+}
+
+/// Stable across protocol versions of the same conversation shape.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ConversationClass {
+    Private,
+    Group,
+}
```
jazzz, line 70: 👌

core/conversations/src/context.rs
```diff
@@ -225,62 +226,44 @@ where
...
         let env = EnvelopeV1::decode(payload)?;
 
         // TODO: Impl Conversation hinting
         let convo_id = env.conversation_hint;
 
         match convo_id {
-            c if c == self.inbox.id() => self.dispatch_to_inbox(&env.payload),
-            c if c == self.pq_inbox.id() => self.dispatch_to_inbox2(&env.payload),
+            c if c == self.inbox.id() => self.dispatch_to_inbox(&env.payload).map(Into::into),
+            c if c == self.pq_inbox.id() => self.dispatch_to_inbox2(&env.payload).map(Into::into),
```
jazzz, line 233: [Dust] Change return type of dispatch functions to handle the conversion into PayloadOutcome. 

This would reduce the clutter at the callsite.

core/conversations/src/outcomes.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub initial: Option<ConvoOutcome>,
+}
+
+#[derive(Debug, Clone, Default)]
+pub enum PayloadOutcome {
+    #[default]
+    Empty,
+    Convo(ConvoOutcome),
+    Inbox(InboxOutcome),
+}
```
jazzz, line 51: [.] In [my original suggestion](https://github.com/logos-messaging/libchat/pull/106#discussion_r3312779925) I was implying to merge the outcomes into a single struct type, to keep with your original design. 

Using an enum works too. Up to you ; just wanted to correct an ambiguity.

jazzz (approved):
The changes so far address all my concerns. Happy to merge when you are done @osmaczko

core/conversations/src/inbound.rs
osmaczko: In the order the conversation's state machine yields them, is better described in the code comments now

core/conversations/src/conversation/group_v1.rs
osmaczko: That's the consequence of `pub type ConversationIdOwned = Arc<str>;`.  I'll get rid of `Arc` usage from within core altogether.

core/conversations/src/inbound.rs
osmaczko: > * Cardinality : Previously each inbound payload could generate at most one message. I'm not opposed to allowing multiple as a future proofing approach, but is there a usecase you can think of where this is needed?

I don't see usecase now and I agree it is premature. Switched back to one message.

> ConvoMultiplexing : Previously each payload was associated to a single conversation. This format allows for FrameOutput to contain different convo_ids. Can you explain why that is needed? It appears to be a significant change to the parsing rules.

Same as above, premature. Removed.

osmaczko: > I'm having a hard time seeing how InboundResult is a notable improvement over
> 
> **ContentData++**
> 
> Am I missing something important?

No, you are not missing anything. The back-and-forth is what muddied it. The core went ContentData -> events -> back to a structured result that landed functionally equivalent to the original. Your dedicated `PayloadOutcome` enum (Empty / Convo / Inbox) is the real improvement, so I've adopted it.

PR #120, opened 2026-05-26
Include sender information on missing messages

### LINKED ISSUE #98: Group participants in a conversation can tell if a message is missing, and who sent it.

kaichaosun:
In previous PR https://github.com/logos-messaging/libchat/pull/105, we implemented causal history for detecting missing message. 
This PR introduce extra `sender_id` to detect who the missing messages belongs to.

Fix https://github.com/logos-messaging/libchat/issues/98

### DIFF
--- a/core/conversations/src/causal_history.rs
+++ b/core/conversations/src/causal_history.rs
@@ -49,7 +49,7 @@ pub struct Frontier {
 }
 
 impl Frontier {
-    /// Derive a fresh `MessageId` for an outbound message.
+    /// Construct a fresh `Frontier` for an outbound message.
     pub fn new(sender_id: String, message_id: String) -> Self {
         Self {
             sender_id,
@@ -120,9 +120,9 @@ struct ConvoState {
 }
 
 impl ConvoState {
-    fn record_seen(&mut self, id: Frontier) {
-        if self.seen.insert(id.clone()) {
-            self.frontiers.push_back(id);
+    fn record_seen(&mut self, info: Frontier) {
+        if self.seen.insert(info.clone()) {
+            self.frontiers.push_back(info);
             while self.frontiers.len() > CAUSAL_HISTORY_LEN {
                 self.frontiers.pop_front();
             }
@@ -174,13 +174,13 @@ impl CausalHistoryStore {
             })
             .collect();
 
-        let wire_id = frontier.encode();
+        let encoded_info = frontier.encode();
         // Our own message joins the seen-set so it appears in our future
         // causal history (and, later, so we can ack peers' references to it).
         state.record_seen(frontier);
 
         ReliablePayload {
-            message_id: wire_id,
+            message_id: encoded_info,
             channel_id: conversation_id.to_owned(),
             lamport_timestamp: lamport,
             causal_history,
@@ -207,11 +207,11 @@ impl CausalHistoryStore {
 
         let mut detected = Vec::new();
         for entry in &payload.causal_history {
-            let id = Frontier::decode(&entry.message_id)?;
-            if !state.seen.contains(&id) && state.reported_missing.insert(id.clone()) {
+            let frontier = Frontier::decode(&entry.message_id)?;
+            if !state.seen.contains(&frontier) && state.reported_missing.insert(frontier.clone()) {
                 let m = MissingMessage {
                     conversation_id: conversation_id.to_owned(),
-                    frontier: id,
+                    frontier: frontier,
                 };
                 detected.push(m.clone());
                 missing.push(m);
@@ -303,24 +303,6 @@ mod tests {
         receiver.on_receive("c", &m1).unwrap();
         receiver.on_receive("c", &m2).unwrap();
         assert!(receiver.take_missing().is_empty());
-    }
-
-    #[test]
-    fn message_id_carries_account_id_verbatim() {
-        let s = CausalHistoryStore::new();
-        let a = payload(&s, "c", "alice", b"1");
-        let b = payload(&s, "c", "alice", b"2");
-        let c = payload(&s, "c", "bob", b"3");
-
-        assert_eq!(
-            Frontier::decode(&a.message_id).unwrap().sender_id(),
-            "alice"
-        );
-        assert_eq!(
-            Frontier::decode(&b.message_id).unwrap().sender_id(),
-            "alice"
-        );
-        assert_eq!(Frontier::decode(&c.message_id).unwrap().sender_id(), "bob");
     }
 
     #[test]
--- a/core/conversations/src/causal_history.rs
+++ b/core/conversations/src/causal_history.rs
@@ -20,15 +20,11 @@
 
 use std::cell::RefCell;
 use std::collections::{HashMap, HashSet, VecDeque};
-use std::fmt;
 use std::rc::Rc;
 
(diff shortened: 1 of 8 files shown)


core/conversations/src/causal_history.rs
```diff
@@ -97,23 +163,24 @@ impl CausalHistoryStore {
...
 
+        let encoded_info = frontier.encode();
         // Our own message joins the seen-set so it appears in our future
         // causal history (and, later, so we can ack peers' references to it).
-        state.record_seen(message_id.clone());
+        state.record_seen(frontier);
 
         ReliablePayload {
-            message_id,
+            message_id: encoded_info,
```
kaichaosun: ~~In previous [discussion](https://www.notion.so/Reliable-Messaging-in-Libchat-3678f96fb65c80a5acd8ff8d10c7539a?d=3688f96fb65c806e9908001c95fd3ee0&source=copy_link#3678f96fb65c809589a4ff2f3ca276c9), we decided to reuse the `message_id` in `ReliablePayload` protobuf for sender info.~~
~~Future improvements includes wire format naming and frontiers calculations is planned later.~~

core/conversations/src/causal_history.rs
```diff
@@ -20,11 +20,77 @@
...
 use std::collections::{HashMap, HashSet, VecDeque};
+use std::fmt;
 use std::rc::Rc;
 
+use crate::errors::ChatError;
 use crate::proto::{Bytes, HistoryEntry, ReliablePayload};
 use crate::utils::{blake2b_hex, hash_size};
 
+/// `hash_size::MessageId` is U32 (32 bytes) → 64 hex chars.
+const MESSAGE_ID_HEX_LEN: usize = 64;
```
jazzz: [Sand] As hash_size are defined at compile time, you can derive constants from them.

```
use blake2::digest::typenum::Unsigned;
const MESSAGE_ID_HEX_LEN: usize = 2 * <hash_size::MessageId as hash_size::HashLen>::Size::USIZE;
```

Deriving it would allow it to be automatically updated should the value change in the future. Thought its not the prettyiest syntax currently. Your call.

core/conversations/src/causal_history.rs
```diff
@@ -20,11 +20,77 @@
...
+        let split = s.len() - MESSAGE_ID_HEX_LEN;
+        Ok(Self {
+            sender_id: s[..split].to_owned(),
+            message_id: s[split..].to_owned(),
+        })
+    }
+
+    /// Render to the wire-form string used by the `message_id` proto field.
+    pub fn encode(&self) -> String {
+        self.to_string()
```
jazzz: [Dust] Encoding as a String seem unnecessarily wasteful. Bytes would be more compact on the wire.

core/conversations/src/causal_history.rs
```diff
@@ -20,11 +20,77 @@
...
+    pub fn decode(s: &str) -> Result<Self, ChatError> {
+        if s.len() < MESSAGE_ID_HEX_LEN {
+            return Err(ChatError::BadParsing("MessageId"));
+        }
+        let split = s.len() - MESSAGE_ID_HEX_LEN;
+        Ok(Self {
+            sender_id: s[..split].to_owned(),
+            message_id: s[split..].to_owned(),
+        })
+    }
```
jazzz: [Sand] Prefer self describing memory layouts, when possible. 

This is more of a design principal - but helpful to keep in mind.

This code as written is tightly coupled to the code that produced it - changing `MESSAGE_ID_HEX_LEN` would be break interop between clients. 

Three improved design choices in order of robustness:
- Deterministic parsing via TLV, such that any recipient can parse the values regardless of how they were generated.
- Use a defined delimiter such as '|' or ':'.  Decouples from payload from size changes but requires consistent order.
- Use a single byte `Sigil` to define which parsing format should be used.

Delimiter seems like the right trade off here, given the string implementation and the narrowed scope.

core/conversations/src/causal_history.rs
```diff
@@ -97,23 +163,24 @@ impl CausalHistoryStore {
...
+            .frontiers
             .iter()
-            .cloned()
-            .map(|message_id| HistoryEntry {
-                message_id,
+            .map(|f| HistoryEntry {
+                message_id: f.encode(),
                 retrieval_hint: Bytes::new(),
             })
             .collect();
```
jazzz: [Dust] consider flattening Frontier to a String with dedicated encode and decode functions. 

I love adding types especially when:
- Types represent a state transition through a pipeline. TypeState pattern.
- Theres complicated functionality being provided
- A Type is expected to grow to take on new behavior.

In this case it seems like we may just want to define causual_history::message_id as a string, which is a combination of the sender_id and derived_message_id. 

Creating a type that is only used to concat two strings together, distracts from what is really happening.

jazzz: Specifically 'fn derived_message_id` already has access to SenderId and could handle the concat in one step.

core/conversations/src/causal_history.rs
```diff
@@ -45,20 +111,20 @@ struct ConvoState {
...
     /// Every message ID delivered locally (own sends + received).
-    seen: HashSet<String>,
+    seen: HashSet<Frontier>,
     /// Bounded frontier of recently-seen IDs (oldest first) attached to
     /// outbound messages as causal history.
-    frontier: VecDeque<String>,
+    frontiers: VecDeque<Frontier>,
     /// Missing IDs already reported, so a gap is surfaced exactly once.
-    reported_missing: HashSet<String>,
+    reported_missing: HashSet<Frontier>,
```
jazzz, line 87: [Dust] I like the removal of naked String types from the code base.

If you decided to flatten Frontier, consider using a type def or NewType wrapper (depending in your vision) to provide the same Type safety/Readability.

jazzz (approved):
Change looks good. 

My comments are all future considerations.

core/conversations/src/causal_history.rs
kaichaosun: I try to make it consistent with [nim-sds](https://github.com/logos-messaging/nim-sds/blob/master/sds/types/sds_message_id.nim#L4). Worth to make a change later with more communication happened.

I do notice that nim-sds add sender id recently, this is protobuf updated: https://github.com/logos-messaging/chat_proto/pull/9, also updates changes here to use new wire format.

core/conversations/src/causal_history.rs
kaichaosun: Not needed in new wire format.

core/conversations/src/causal_history.rs
kaichaosun: Looks like the Frontier struct has more meaning after the new wire format.
`derived_message_id ` is isolated as we may want to introduce a consistent reusable function to calculate the identifier going through libchat.

osmaczko (approved):

PR #123, opened 2026-05-28
refactor(core): replace Rc-based Context with a synchronous, Send-able Core

osmaczko:
Make the conversations core Send so the threaded client can own it behind an
Arc<Mutex<Core>>: a background worker polls the transport and handles inbound
payloads while the application thread issues outbound calls (send, create
conversation). Sharing the core across those two threads means moving it into
the spawned worker, which is only legal if it is Send. Access stays serialized
by the client's Mutex (one thread at a time), so the core needs Send but not
Sync and carries no lock of its own. See
docs/adr/0001-client-event-system.md for the background-poller design.

The Rc<RefCell> service-sharing is what made the core !Send. Context is de-Rc'd
and renamed to Core, owning its services outright and driving the inbox and
conversation primitives with plain &mut self.

- Services (identity, delivery, store, registry, MLS context, causal history)
  are bundled into a ServiceContext\<S\> behind an ExternalServices trait, with
  S = (DS, RS, CS). Constructors live on the (DS, RS, CS) form because S cannot
  be inferred backwards through S::DS.
- Inbox, InboxV2, PrivateV1Convo, and GroupV1Convo become non-generic and
  receive the ServiceContext bundle as a &mut/& parameter; no Rc or
  RefCell-as-shared-state remains, so Core is Send whenever its injected
  services are.
- Dispatch branches on ConversationKind in one place: Core rebuilds the target
  as a Convo\<S\>/GroupConvo\<S\> trait object bound to the service bundle, so
  conversations never escape the orchestrator.
- CausalHistoryStore drops its Rc, keeping a plain RefCell.


jazzz:
@osmaczko can you link to the project issue that this PR belongs to?

osmaczko:
> @osmaczko can you link to the project issue that this PR belongs to?

This is part of [Client Event System](https://github.com/logos-messaging/libchat/issues/97), as per [ADR's 0001](https://github.com/logos-messaging/libchat/blob/2d3ad27d518ed2dbfe8cd305eb9330bb2e34f19c/docs/adr/0001-client-event-system.md):

> Event is an asynchronous notification. The client's constructor returns a Receiver<Event> alongside the client handle. A background poller drives the transport, calls into the core for each inbound payload, translates the resulting PayloadOutcome into one event per observation, and pushes them onto the channel. Background work that has no synchronous trigger at all (delivery retry timeouts, future protocol timers) pushes onto the same channel.

This PR is a prerequisite for enabling the client to perform background, threaded transport polling. I split the changes into two PRs (this + followup with background processing) to make the review easier.

Do you think it makes sense to close Client Event System issue and open new one to track that? For instance: _Threaded Client_?

core/conversations/src/core.rs
```diff
@@ -0,0 +1,373 @@
...
+
+// This is the main entry point to the conversations api.
+// `Core` manages lifetimes of objects to process and generate payloads.
+//
+// Fully synchronous and single-threaded: it owns its services outright (no
+// interior mutability, no shared ownership) and drives the inbox/conversation
+// primitives with plain `&mut self`.
+pub struct Core<DS: DeliveryService, RS: RegistrationService, CS: ChatStore> {
+    identity: Identity,
+    ds: DS,
```
jazzz: [!] Longterm Its uncertain if the core can be the owner of the DS. There are possible usecases where the client, or the app may require access to the DS.

crates/client/src/client.rs
```diff
@@ -32,68 +32,54 @@ impl<D: DeliveryService + 'static> ChatClient<D> {
...
-    ) -> Result<ConversationId, ClientError<D::Error>> {
+    ) -> Result<ConversationId, ClientError> {
         let intro = Introduction::try_from(intro_bundle)?;
-        let (convo_id, envelopes) = self.ctx.create_private_convo(&intro, initial_content)?;
-        self.dispatch_all(envelopes)?;
-        Ok(convo_id)
+        self.core
+            .create_private_convo(&intro, initial_content)
+            .map_err(Into::into)
     }
```
jazzz, line 93: This is much cleaner. Love this direction.

crates/client/src/client.rs
```diff
@@ -32,68 +32,54 @@ impl<D: DeliveryService + 'static> ChatClient<D> {
...
-        convo_id: &str,
-        content: &[u8],
-    ) -> Result<(), ClientError<D::Error>> {
-        let envelopes = self.ctx.send_content(convo_id, content)?;
-        self.dispatch_all(envelopes)
+    /// Encrypt and send `content` to an existing conversation.
+    pub fn send_message(&mut self, convo_id: &str, content: &[u8]) -> Result<(), ClientError> {
+        self.core
+            .send_content(convo_id, content)
+            .map_err(Into::into)
```
jazzz, line 104: [Dust] The map_error can be avoided, by invoking the conversion directly.

as     `Chat(#[from] ChatError),` is defined; You can use the try operator with an explicit Ok of unit

```
  self.core.send_content(convo_id, content)?;
  Ok(())
```

core/conversations/src/core.rs
```diff
@@ -0,0 +1,373 @@
...
+
+    /// Add members to an existing group conversation.
+    pub fn group_add_member(
+        &mut self,
+        convo_id: &str,
+        members: &[&AccountId],
+    ) -> Result<(), ChatError> {
+        let mut convo = self.load_mls_convo(convo_id)?;
+        convo.add_member(members, &self.mls_ctx, &mut self.ds, &self.contact_registry)
+    }
```
jazzz, line 207: [Sand] By removing the GroupConvoTrait, this now becomes a runtime error (rather than compile time) when account_id is not of the correct type. 

Not critical for V0.2 however less than Ideal.

jazzz: [?] What was the issue that led to the removal of the ConvoTraits Traits

core/conversations/src/core.rs
```diff
@@ -0,0 +1,373 @@
...
+    /// Encrypt and publish `content` to an existing conversation.
+    pub fn send_content(&mut self, convo_id: &str, content: &[u8]) -> Result<(), ChatError> {
+        let record = self.load_conversation_meta(convo_id)?;
+        match record.kind {
+            ConversationKind::PrivateV1 => self.private_send_content(record, content),
+            ConversationKind::GroupV1 => self.group_send_content(&record.local_convo_id, content),
+            ConversationKind::Unknown(_) => Err(ChatError::BadBundleValue(format!(
+                "unsupported conversation type: {}",
+                record.kind.as_str()
+            ))),
```
jazzz: [Dust] Harder to manage over dynamic dispatch

core/conversations/src/core.rs
```diff
@@ -0,0 +1,373 @@
...
+#[allow(dead_code)]
+fn assert_core_is_send<DS, RS, CS>()
+where
+    DS: DeliveryService + Send + 'static,
+    RS: RegistrationService + Send + 'static,
+    CS: ChatStore + Send + 'static,
+{
+    fn is_send<T: Send>() {}
+    is_send::<Core<DS, RS, CS>>();
+}
```
jazzz: [Sand] This is cute - However It doesn't seem needed.

1) If you want to enforce that DS, RS, CS are Send so that Send can be auto implemented on Core
   - Add the Send requirement to service_traits::DeliveryService. That way its enforced everywhere cleanly

2) Better yet. Core does not require Send to operate. It's only needed if Client requires Send. Move this constraint to the Client. Have the client enforce that all types that are passed int are Send compatible. This enforce the compile time check when building Client, and allows test cases or other use cases to use !Send Services if there is no conflict.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -363,29 +319,18 @@ where
...
-    fn add_member(&mut self, members: &[&AccountId]) -> Result<(), ChatError> {
-        let ctx_ref = self.ctx.borrow();
-        let provider = ctx_ref.provider();
+    pub fn add_member<MlsCtx: MlsContext, DS: DeliveryService, KP: KeyPackageProvider>(
+        &mut self,
+        members: &[&AccountId],
+        ctx: &MlsCtx,
+        ds: &mut DS,
+        keypkg_provider: &KP,
+    ) -> Result<(), ChatError> {
```
jazzz, line 278: Moving services to be passed in on invocation is a positive change; rather than references being retained which complicates ownership.

Specifically as this PR removes dynamic dispatch, there are no requirements for functions to be similar across Conversations of the the same class.

However

[Sand] Consider combining Services into a common provider object. There are more services required to facilitate DeMLS. Adding them all as parameters will be harder to manage at various callsites, and implementations. 

Here is an example usage:[ group_v1::add_client](https://github.com/logos-messaging/libchat/blob/main_mls/core/core_client/src/conversation/group_v1.rs#L240-L244)

With Core mapping to [core_client::InnerClient](https://github.com/logos-messaging/libchat/blob/main_mls/core/core_client/src/core_client.rs#L139-L147)

If desirable to constrain which services a function has access to, the ExternalServices bounds can be updated per function.

jazzz (commented):
In general. I like this direction. 

- Removing run time borrow checks is absolutely needed.
- Renaming Context is a welcome change.

I'm not entirely convinced that removing dynamic dispatch is needed/helpful in the long term. But if it simplifies things now, I don't see any blockers.

core/conversations/src/core.rs
osmaczko: Legitimate forward-looking flag :+1: If that ever becomes the case, shared ownership should be introduced. Let’s KISS for now.

jazzz: > shared ownership should be introduced

Shared ownership is complicated. I'd look at passing in the ProviderContext to Core by reference. That way it could be owned by client, without any runtime borrow checks

osmaczko:
@jazzz, I’ve taken your feedback into consideration and went with combining the services `ServicesContext` I hope this makes your future de'mls merge easier. Please re-review.

core/conversations/src/core.rs
osmaczko: I agree, will move this part to the Client.

core/conversations/src/core.rs
osmaczko: Brought back GroupConvoTrait.

core/conversations/src/conversation/group_v1.rs
osmaczko: I've combined Services into common `ServiceContext`.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -2,153 +2,99 @@
...
-            keypkg_provider,
             mls_group,
             convo_id,
-            causal,
         })
     }
 
     // Configure the delivery service to listen for the required delivery addresses.
-    fn subscribe(ds: &mut DS, convo_id: &str) -> Result<(), ChatError> {
+    fn subscribe(ds: &mut impl DeliveryService, convo_id: &str) -> Result<(), ChatError> {
```
jazzz, line 97: [Sand] Using a trait impl adds build complexity unnecessarily. Use the generic type S::DS

core/conversations/src/core.rs
```diff
@@ -0,0 +1,321 @@
...
+pub use crate::conversation::ConversationId;
+pub use crate::inbox::Introduction;
+
+// This is the main entry point to the conversations api.
+// `Core` manages lifetimes of objects to process and generate payloads.
+//
+// Fully synchronous and single-threaded: it owns its services outright (no
+// interior mutability, no shared ownership) and drives the inbox/conversation
+// primitives with plain `&mut self`.
+pub struct Core<S: ExternalServices> {
```
jazzz, line 27: [Dust] Consider more verbose name. `Core` doesn't provide much detail. E.g. CoreClient, CoreEngine, ChatCore

core/conversations/src/inbox_v2.rs
```diff
@@ -53,59 +44,31 @@ pub trait MlsProvider: OpenMlsProvider {
...
-{
+pub struct InboxV2 {
     // Account_id field is an owned value, so it can be returned via reference.
     account_id: AccountId,
-    account: Rc<RefCell<MlsIdentityProvider<IP>>>,
-    ds: Rc<RefCell<DS>>,
-    reg_service: Rc<RefCell<RS>>,
-    store: Rc<RefCell<CS>>,
-    causal: CausalHistoryStore,
-    mls_provider: Rc<RefCell<MlsEphemeralPqProvider>>,
```
jazzz, line 67: 🙏

core/conversations/src/inbox_v2.rs
```diff
@@ -117,19 +80,11 @@ where
...
-            self.ds.clone(),
-            self.reg_service.clone(),
-            self.causal.clone(),
-        )
-    }
-
-    pub fn handle_frame(&self, payload_bytes: &[u8]) -> Result<InboxOutcome, ChatError> {
+        payload_bytes: &[u8],
+        cx: &mut ServiceContext<S>,
+    ) -> Result<InboxOutcome, ChatError> {
```
jazzz, line 87: In GroupV1Convo the function argument order is `Self, ServiceContext, args`

[Sand] Consider using that order everywhere for consistency. 

Arguments order should be based on variability.

core/conversations/src/lib.rs
```diff
@@ -16,12 +17,12 @@ pub use account::LogosAccount;
...
 pub use chat_sqlite::ChatStorage;
 pub use chat_sqlite::StorageConfig;
-pub use context::{Context, ConversationId, Introduction};
-pub use conversation::GroupConvo;
+pub use core::{ConversationId, Core, Introduction};
 pub use errors::ChatError;
 pub use outcomes::{
     Content, ConversationClass, ConvoOutcome, InboxOutcome, NewConversation, PayloadOutcome,
 };
+pub use service_context::ExternalServices;
```
jazzz, line 25: [?] This is an internal representation isn't it? Does it need to be exported from the crate?

core/conversations/src/service_context.rs
```diff
@@ -0,0 +1,98 @@
...
+    type CS = CS;
+}
+
+/// Bundles every service a conversation operation may need.
+pub(crate) struct ServiceContext<S: ExternalServices> {
+    pub(crate) ds: S::DS,
+    pub(crate) registry: S::RS,
+    pub(crate) store: S::CS,
+    pub(crate) mls_identity: MlsIdentityProvider<LogosAccount>,
+    pub(crate) mls_provider: MlsEphemeralPqProvider,
```
jazzz, line 36: [Future] From a maintenance perspective the MLS details probably should be contained to a limited scope. They aren't External services strictly speaking, and represent an internal state.

In the future a pattern which handles both external services and shared conversation state should be explored.

core/conversations/src/core.rs
```diff
@@ -0,0 +1,321 @@
...
+    services: ServiceContext<S>,
+    inbox: Inbox,
+    pq_inbox: InboxV2,
+}
+
+// Constructors live on the `(DS, RS, CS)` form: `S` can't be inferred backwards
+// through `S::DS`, so the bundle is built from the three args here.
+impl<DS, RS, CS> Core<(DS, RS, CS)>
+where
```
jazzz, line 36: [Pebble] Exposing the tuple seems unnecessary (perhaps I'm missing something). This has impact on every app and test case and should be used with caution. 

I think you wanted to implement ExternalServices for the tuple so that It can be built internally.

`impl<DS, RS, CS> ExternalServices for (DS, RS, CS) ... { ...`

core/conversations/src/conversation.rs
```diff
@@ -1,47 +1,39 @@
...
-    fn id(&self) -> &str;
-}
-
-pub trait Convo: Id + Debug {
-    fn send_message(&mut self, content: &[u8])
-    -> Result<Vec<AddressedEncryptedPayload>, ChatError>;
+/// Behaviour shared by every conversation kind.
+pub(crate) trait Convo<S: ExternalServices> {
+    fn send_content(&mut self, cx: &mut ServiceContext<S>, content: &[u8])
+    -> Result<(), ChatError>;
```
jazzz, line 18: This is a cleaner approach, and is much welcome

jazzz (approved):
This approach looks good to me. 

Theres some improvements which could be made, but from a lean design approach I think this solves the problem nicely. 

- Provides flexibility while its needs the most.
- Approach is self contained within the crate; Clear separation of concerns. 
- Removes complex runtime borrow checks. 

I don't see any blockers, and follow up tasks can be completed later if desired

osmaczko:
> This approach looks good to me.
> 
> Theres some improvements which could be made, but from a lean design approach I think this solves the problem nicely.
> 
> * Provides flexibility while its needs the most.
> * Approach is self contained within the crate; Clear separation of concerns.
> * Removes complex runtime borrow checks.
> 
> I don't see any blockers, and follow up tasks can be completed later if desired

Thanks for review :pray: I'll address comments in a followup PR. Merging this one to avoid yet another conflict-resolution.

PR #124, opened 2026-06-01
feat: http server based key package registry

### LINKED ISSUE #110: Testnet KeyPackage Registry
## Problem

MLS KeyPackages contain the required material to add a new member to a GroupChat. Passing KeyPackages out of band is not viable given the size of ML-KEM keys.

There currently does not exist a way for a Group Participant to add another by account address. 

In the future this will be perform by a robust service likely using λLEZ. However there is more careful thought required to execute on that.

## Whats needed

There is a current need for a Service which provides these KeyBundles. The service will be eventually replaced in V0.3, so it only needs to provide the bare minimum in functionality. 

- [x] LocalIdentities can submit KeyPackages to the service
- [x] Clients can fetch a KeyPackage for a given LocalIdentity.
- [x] LocalIdentities can submit new Bundles.
- [x] Service stores the latest N KeyPackages for a given LocalIdentity
- [x] Service only accepts KeyBundles from the the correct LocalIdentity.
- [x] Service retains KeyBundles for N days. 

## Suggestions
- Implement KeyPackage submission via LogosDelivery. Clients publish to a fixed topic, All Service instances listen.
- Implement Fetch however is easiest.

kaichaosun:
Changes:
- Adds a standalone HTTP service bin/keypackage-registry that holds MLS KeyPackages so clients can add a contact by account_id without out-of-band bundle exchange.
- Adds a matching HttpRegistry client in extensions/components and an opt-in `--registry-url` flag on chat-cli. Without the flag, behavior is unchanged.
- Throwaway by design: scheduled to be replaced by λLEZ-based discovery in v0.3.

Fix https://github.com/logos-messaging/libchat/issues/110.


extensions/components/src/lib.rs
```diff
@@ -1,7 +1,9 @@
 mod contact_registry;
 mod delivery;
+mod http_registry;
 mod storage;
 
 pub use contact_registry::EphemeralRegistry;
 pub use delivery::*;
+pub use http_registry::{HttpRegistry, HttpRegistryError};
```
jazzz: [Sand] To maintain clean imports/organization, This type should be created within the contact_registry module.

bin/keypackage-registry/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+## API
+
+### `POST /v0/keypackage`
+
+Submit a bundle.
+
+```json
+{
+  "account_id":    "user-chosen label",
```
jazzz: [Pebble] Account Id cannot be a user-chosen label. It must be a static, string derived from the account

bin/keypackage-registry/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+The storage schema is multi-device-ready — an `account_id` can have several `device_pubkey`s, each with its own keypackage history. The current chat layer (Scope A) only consumes one device per account, so `GET` returns the single latest bundle across all devices.
+
+**Server is dumb storage.** Submissions are stored without verification. Bundles carry a self-signature by the device key over `(account_id || device_pubkey || key_package || ts_le)`; consumers verify it on retrieve. This catches a replay attack where someone copies a victim's bundle and re-posts it under a different `account_id` — the signature commits to the original `account_id` and won't verify against the new one.
+
+The self-signature does **not** prove that `device_pubkey` is actually authorized for `account_id` (an attacker can mint their own keys and sign anything). That impersonation gap closes once λLEZ provides the on-chain authorization mapping in v0.3.
```
jazzz: The problem here is that this registry is attempting to solve two problems at once.

Account registration (The association of Installations/DeviceKeys into a single account) and the caching of KeyPackages from a single DeviceKey. 

Solving both at the same time adds more complication. I'd focus only on the KeyPackage caching for now, then address Accounts in a later PR.

bin/keypackage-registry/README.md
```diff
@@ -0,0 +1,119 @@
...
+  "account_id":    "user-chosen label",
+  "device_pubkey": "base64(32-byte ed25519 verifying key)",
+  "key_package":   "base64(MLS KeyPackage bytes)",
+  "timestamp_ms":  1717200000000,
+  "signature":     "base64(64-byte ed25519 signature)"
+}
+```
+
+`signature` MUST be Ed25519 by `device_pubkey` over the byte concatenation
+`account_id || device_pubkey || key_package || timestamp_ms.to_le_bytes()`.
```
jazzz: [Dust] Having the signature being defined over byte concat, and the "Package" being transported to the server being defined as base64 encoded JSON, makes validating the payload more difficult. 

Conside symmetry, by using a well defined binary payload, with an associated signature.

bin/keypackage-registry/README.md
```diff
@@ -0,0 +1,119 @@
...
+## API
+
+### `POST /v0/keypackage`
+
+Submit a bundle.
+
+```json
+{
+  "account_id":    "user-chosen label",
+  "device_pubkey": "base64(32-byte ed25519 verifying key)",
```
jazzz: [?] What key does this correspond to?

bin/keypackage-registry/README.md
```diff
@@ -0,0 +1,119 @@
...
+}
+```
+
+`signature` MUST be Ed25519 by `device_pubkey` over the byte concatenation
+`account_id || device_pubkey || key_package || timestamp_ms.to_le_bytes()`.
+
+The server validates only basic shapes (`device_pubkey` 32 bytes, `signature`
+64 bytes, valid base64). It does **not** verify the signature — verification
+happens client-side on retrieve. Returns `204 No Content` on success, `400`
+on shape errors.
```
jazzz: [?] Is there a reason why the serve couldn't validate signatures in this test case?

bin/keypackage-registry/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+### `GET /v0/keypackage/{account_id}`
+
+Returns the most recently submitted bundle for that id, across all devices.
+
```
jazzz: [?] What is considered a bundle in this case?

jazzz: The issue with querying by account_id is that key-packages are single use, per group. If one of the keypackages have already been used, its harder to request another another one.

bin/keypackage-registry/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "keypackage-registry"
+version = "0.1.0"
+edition = "2024"
```
jazzz, line 4: This will need need to be deployed and managed in its own repository.

bin/keypackage-registry/README.md
kaichaosun: The server could validate the signature in this test phrase, but when we switch to LEZ or shared key package store with many such servers, this assumption may change. 
Client side validation seems always needed, probably can be omitted if we have a trusted layer for such validation like blockchain, LEZ could potentially validate ed25519 signature, but I'm not sure about if such primitives is available for dapps or not.

bin/keypackage-registry/README.md
kaichaosun: It's the `verifying_key` in `LogosAccount`, not sure what's the best concept we want involve here.

bin/keypackage-registry/README.md
kaichaosun: @jazzz  The retrieve api is currently use account id, should I change it to use signing/device pubkey instead?

https://github.com/logos-messaging/libchat/blob/2d3ad27d518ed2dbfe8cd305eb9330bb2e34f19c/extensions/components/src/contact_registry.rs#L68

jazzz: @kaichaosun. I think so. Its a cleaner model for key rotation. 

`AccountId --1:M--> DeviceIdentity --1:M--> KeyPackage`

For now I'd create a type def to a String, and we can easily replace it after.  

I can see the KeyPackage provider trait eventually being move to the client level, but that would be future work

bin/keypackage-registry/README.md
jazzz: Yeah LogosAccount will need some work - Specifically it has no concept of a local DeviceIdentity. I'll draft the beginning of a spec for discussion.

bin/keypackage-registry/README.md
jazzz: > Client side validation seems always needed, 

Absolutely. This is a trustless system, clients always need to validate. 

> The server could validate the signature in this test phrase,

If its not much work, having serverside validation would be helpful to avoid a DOS where an adversary submits excessive invalid bundles. This is outside of the security model for the time being, so not critical. But worthwhile if relatively easy. 

Clients would always need to validate the signature, to hold the service accountable - however the service can hold submitters accountable by verifying the bundles are valid before persisting them.

bin/chat-cli/src/main.rs
```diff
@@ -104,18 +110,38 @@ fn run<D: DeliveryService + 'static>(
...
+            launch_tui(client, inbound, cli)
+        }
+        None => {
+            let client = ChatClient::open(cli.name.clone(), storage, transport)
+                .map_err(|e| anyhow::anyhow!("{e:?}"))
+                .context("failed to open chat client")?;
+            launch_tui(client, inbound, cli)
+        }
+    }
+}
```
jazzz, line 134: I wouldn't be afraid to be opinionated here. Too Much configuration makes it harder to use. 

[Sand] Consider forcing HTTPRegistry with a fixed URL.

bin/chat-cli/src/main.rs
```diff
@@ -104,18 +110,38 @@ fn run<D: DeliveryService + 'static>(
...
+        key: "chat-cli".to_string(),
+    };
+
+    match cli.registry_url.as_deref() {
+        Some(url) => {
+            let registry = HttpRegistry::new(url);
+            let client =
+                ChatClient::open_with_registry(cli.name.clone(), storage, transport, registry)
+                    .map_err(|e| anyhow::anyhow!("{e:?}"))
+                    .context("failed to open chat client with HTTP registry")?;
```
jazzz, line 124: [Sand] Mostly unrelated to this PR. However I think we should consider making a ChatClient opinionated or Create a LogosChatClient, which is default prefconfigured with Logos Specific services. 

Every configuration has the potential for clients to fragment. Using defined settings, would help developers maintain interop, with less maintenance. 

`LogosChat = Client<LogosCoreDelivery, LogosAccounts, HTTP Keypackage Registry... etc.>`

jazzz (approved):
I think these are good changes. 

- Great use of the extensions to store configurable service implementations. 

High Level, for future PRs we'd like to see:

- Chat-cli become leaner and simpler.
- Developer Choice + Sensible defaults: Its good to give developers options to configure, but we should also have a default code path, that is simple and straight forward to use.


Once AccountId has been decoupled from DeviceIdentity, I'd be happy to merge.

bin/chat-cli/src/main.rs
kaichaosun: The idea sounds great, created an issue https://github.com/logos-messaging/libchat/issues/126

bin/chat-cli/src/main.rs
kaichaosun: Once we have deployed "prod" contact-store service (for http restristration), we can add the fixed url then.

bin/keypackage-registry/Cargo.toml
kaichaosun: I have created repo https://github.com/logos-messaging/chat-store, but hesitate to move this crate to this repo when performing, because I think the monorepo will help speed up the development.

I will the leave that repo empty, until we are decided to move this crate out.

bin/keypackage-registry/README.md
kaichaosun: Changed bundle to key package, also replace account id with device id now.

PR #125, opened 2026-06-02
feat(client): add threaded transport polling

osmaczko:
The client, not the app, now drives the transport; events are delivered
asynchronously, per ADR 0001.

- ChatClient owns Arc<Mutex<Core>> + a worker thread.
- The worker select!s over the inbound and shutdown channels; Drop joins it.
  Outbound runs on the caller's thread.
- A single Transport (DeliveryService + inbound()) owns both directions of the
  boundary, so the client takes one transport rather than a (delivery, inbound)
  pair. InProcessDelivery::new, CDelivery, and chat-cli's transports implement it.
- FFI replaces client_receive with client_push_inbound + client_poll_events.
- chat-cli drains Receiver<Event>; inbound and event channels are both crossbeam.
- Corrects ADR 0001's inbound sequence to push — the worker parks on select!,
  it never polls.

### DIFF
--- a/bin/chat-cli/src/transport/file.rs
+++ b/bin/chat-cli/src/transport/file.rs
@@ -2,11 +2,11 @@ use std::collections::BTreeMap;
 use std::fs::{self, File, OpenOptions};
 use std::io::{self, BufReader, Read, Seek, SeekFrom, Write};
 use std::path::{Path, PathBuf};
-use std::sync::mpsc;
 use std::thread;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-use logos_chat::{AddressedEnvelope, DeliveryService};
+use crossbeam_channel::{Receiver, Sender, bounded};
+use logos_chat::{AddressedEnvelope, DeliveryService, Transport};
 
 #[derive(Debug, thiserror::Error)]
 pub enum FileTransportError {
@@ -17,31 +17,31 @@ pub enum FileTransportError {
 #[derive(Debug)]
 pub struct FileTransport {
     transport_dir: PathBuf,
+    inbound_rx: Option<Receiver<Vec<u8>>>,
 }
 
 impl FileTransport {
     /// All instances pointing at the same `transport_dir` share one broadcast bus.
     ///
     /// Messages are written to `{transport_dir}/{delivery_address}/{hours_since_epoch}.bin`
-    /// as length-prefixed frames (`[u32 BE length][payload bytes]`). The background
-    /// thread reads all files under `transport_dir` and forwards every frame to
-    /// the returned channel; `client.receive()` discards frames it cannot decrypt.
-    pub fn new(transport_dir: &Path) -> io::Result<(Self, mpsc::Receiver<Vec<u8>>)> {
+    /// as length-prefixed frames (`[u32 BE length][payload bytes]`). A background
+    /// thread reads all files under `transport_dir` and forwards every frame to the
+    /// inbound stream the client drains via [`Transport::inbound`] (discarding frames
+    /// it cannot decrypt).
+    pub fn new(transport_dir: &Path) -> io::Result<Self> {
         fs::create_dir_all(transport_dir)?;
 
-        let (tx, rx) = mpsc::sync_channel(1024);
+        let (tx, rx) = bounded(1024);
         let dir = transport_dir.to_path_buf();
 
         thread::Builder::new()
             .name("file-transport".into())
             .spawn(move || poll_reader(dir, tx))?;
 
-        Ok((
-            Self {
-                transport_dir: transport_dir.to_path_buf(),
-            },
-            rx,
-        ))
+        Ok(Self {
+            transport_dir: transport_dir.to_path_buf(),
+            inbound_rx: Some(rx),
+        })
     }
 }
 
@@ -68,6 +68,14 @@ impl DeliveryService for FileTransport {
     }
 }
 
+impl Transport for FileTransport {
+    fn inbound(&mut self) -> Receiver<Vec<u8>> {
+        self.inbound_rx
+            .take()
+            .expect("FileTransport::inbound called more than once")
+    }
+}
+
 /// Hours since Unix epoch — used as the rolling filename.
 fn current_hour() -> u64 {
     SystemTime::now()
@@ -77,7 +85,7 @@ fn current_hour() -> u64 {
         / 3600
 }
 
-fn poll_reader(transport_dir: PathBuf, tx: mpsc::SyncSender<Vec<u8>>) {
+fn poll_reader(transport_dir: PathBuf, tx: Sender<Vec<u8>>) {
     // Maps absolute file path → number of bytes already consumed.
     let mut offsets: BTreeMap<PathBuf, u64> = BTreeMap::new();
 
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -1,31 +1,65 @@
 use std::sync::Arc;
+use std::thread::{self, JoinHandle};
 
+use components::EphemeralRegistry;
+use crossbeam_channel::{Receiver, Sender, select};
 use libchat::{
     ChatError, ChatStorage, ConversationId, ConvoOutcome, Core, DeliveryService, InboxOutcome,
     Introduction, PayloadOutcome, RegistrationService, StorageConfig,
 };
-
-use components::EphemeralRegistry;
 use logos_account::TestLogosAccount;
+use parking_lot::Mutex;
 
 use crate::errors::ClientError;
 use crate::event::Event;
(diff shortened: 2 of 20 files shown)


crates/client/src/delivery_in_process.rs
```diff
@@ -1,116 +1,71 @@
...
-            address: address.to_string(),
-            pos,
+    fn publish(&self, address: &str, data: Message) {
+        if let Some(senders) = self.routes.lock().unwrap().get(address) {
+            for tx in senders {
+                // A disconnected endpoint (dropped client) is harmless here.
+                let _ = tx.send(data.clone());
+            }
         }
     }
```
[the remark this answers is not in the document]
jazzz: @osmaczko The DeliveryService trait can be updated if you'd like to return a Result from publish

crates/client/src/client.rs
```diff
@@ -1,27 +1,54 @@
...
-        let store = ChatStorage::in_memory();
-        Self {
-            core: Core::new_with_name(name, delivery, registry, store).unwrap(),
-        }
+    ///
+    /// `inbound` is the transport's inbound channel; the worker blocks on it.
+    pub fn new(
+        name: impl Into<String>,
+        delivery: D,
+        inbound: Receiver<Vec<u8>>,
```
jazzz: [?] I don't understand this.

The DS ownership is being passed to ChatClient, however population of the queue is occurring externally?

crates/client/src/client.rs
```diff
@@ -32,57 +59,130 @@ impl<D: DeliveryService + 'static> ChatClient<D> {
         name: impl Into<String>,
         config: StorageConfig,
         delivery: D,
-    ) -> Result<Self, ClientError> {
+        inbound: Receiver<Vec<u8>>,
+    ) -> Result<(Self, mpsc::Receiver<Event>), ClientError> {
```
jazzz: [Sand] Clients must pass in a crossbeam channel, however receive a std::sync::mpsc Receiver. Is there a reason to not use Crossbeam consistently - The dependency is already required.

jazzz: [Sand] Returning a "Processing Queue" is not very ergonomic for Applications.

We should have an explicit reason why this design choice was made over using callbacks.

crates/client/src/client.rs
```diff
@@ -32,57 +59,130 @@ impl<D: DeliveryService + 'static> ChatClient<D> {
...
+
+    fn spawn(core: ClientCore<D>, inbound: Receiver<Vec<u8>>) -> (Self, mpsc::Receiver<Event>) {
+        let core = Arc::new(Mutex::new(core));
+        let (event_tx, event_rx) = mpsc::channel();
+        let (shutdown_tx, shutdown_rx) = crossbeam_channel::bounded::<()>(0);
+
+        let worker = thread::spawn({
+            let core = Arc::clone(&core);
+            move || worker_loop(core, inbound, shutdown_rx, event_tx)
+        });
```
jazzz, line 122: [Sand] The threading model is not very flexible.

1) LogosDelivery/FileTransport spawn a thread:
    - which places messages into a queue via callback.
2) Client spawns a thread to:
    - listen to the queue
    - processes the result
    - places event into the event queue.
3) App then blocks the mainthread to:
    - poll the events queue
    - update the UI. 


Callbacks from the Client->App would allow for more flexible threading models.e.g AsyncTasks, Direct Invocation, and dedicated UI mainthread. As it stands now, any UI framework would require the blocking thread to listen for messages, and any tokio based system would have to create a blocking thread pool via spawn_blocking

crates/client/src/client.rs
```diff
@@ -32,57 +59,130 @@ impl<D: DeliveryService + 'static> ChatClient<D> {
...
+/// the thread until one of the channels is ready.
+fn worker_loop<D: DeliveryService + Send + 'static>(
+    core: Arc<Mutex<ClientCore<D>>>,
+    inbound: Receiver<Vec<u8>>,
+    shutdown: Receiver<()>,
+    event_tx: mpsc::Sender<Event>,
+) {
+    loop {
+        select! {
+            recv(inbound) -> msg => {
```
jazzz, line 198: [!] DeMLS adds more requirements such as a WakeupService which can call the Conversation after N seconds, to continue processing.

crates/client/src/client.rs
jazzz: This input parameter changes the contract for the DS at the Client level.

[Sand] Be explicit about the requirements from a service. 

Extensions/components/local_broadcaster does not use queues, and would be incompatible with ChatClient. This puts more pressure on App Developers to wire components together. Either add a trait to make an explicit requirement on Services such as DS, or find a more flexible mechanism.

e.g. 
```
trait QueuedResult {
    fn get_event_queue(...)
}
```
or 

```
trait AcceptCallback {
   fn register_callback(...)
}
```

jazzz (commented):
I was expecting to see an AsyncTask model for handling concurrency, however I think the threading model has some strengths too. Specifically it does not require `await`s all over the place.

I think the Queue approach is a huge hit to the Public Developer API, which would add complexity to integrating with existing applications. Not technically a blocker, but it represents a major shift. 

Moving forward I'd like to see a Developer interface emerge which is easier to use, rather than more powerful. This current step appears to be opinionated, however doesn't make ChatClient easier to use. 

I suggest as this work progresses that wither ClientCore becomes signficantly more opinionated. Either Client internally defines Services such as LogosDelivery, and a GenericClient is created later). Or a LogosChatClient is created which is more straight forward and easier to use.  


Happy to approve once there are responses to comments

crates/client/src/client.rs
osmaczko: That's a legitimate concern. The main reason is that DS is defined by Core, which is outbound-only (caller-driven). Core should never call fetch on DS, since that would require a separate thread, nor should it register a callback, which would break its simple, caller-driven model.

The solution I came up with is a `Transport` trait defined by Client:
```rust
pub trait Transport: DeliveryService + Send + 'static {
  fn inbound(&mut self) -> Receiver<Vec<u8>>;
}
```
Transport is now a bidirectional boundary, so the client constructor takes a single parameter.

crates/client/src/client.rs
osmaczko: Fixed. The event channel is crossbeam now too.

osmaczko: The queue is intentional, and it's the more flexible primitive here, not less.

A callback fires on the worker thread, so the app still has to marshal onto its own UI/async context.

Ergonomics belong in the LogosChatClient facade: async on Tokio, Qt signals, or callbacks, all adapters over the pulled events. The generic client stays the minimal mechanism.

This motivated me to write up the target architecture proposal (https://github.com/logos-messaging/libchat/pull/130); we can move the discussion there.

crates/client/src/client.rs
osmaczko: The contract is explicit now: [Transport](https://github.com/logos-messaging/libchat/pull/125#discussion_r3383083234)

crates/client/src/client.rs
osmaczko: A `WakeupService` would be one more producer and one more arm on the worker's select! :thinking:

crates/client/src/client.rs
osmaczko: 1. That thread is the IO itself. logos-delivery is push-only: the node hands payloads to an event callback on its own thread, and there is no fetch the client could drive instead. It exists in any design and belongs to the transport. (FileTransport polls because a filesystem pushes nothing; same conclusion.)
2. That one is deliberate, and it's the only thread the client adds. Decrypt and storage have to run somewhere that is neither the node's callback thread nor the app's. With Client->App callbacks it doesn't go away; it would additionally run app code.
3. The main-thread cost is the same in both models: UI updates run there either way. Polling only adds wake cadence, and the fix for that is a wakeup over the queue: an adapter that parks on recv() and posts into the UI loop, a forwarder into a tokio mpsc (one thread, written once in LogosChatClient, no spawn_blocking per app), or a doorbell callback on the client (notify only, events still pulled) if we want zero extra threads.

jazzz (approved):

bin/chat-cli/src/transport/file.rs
```diff
@@ -17,31 +17,31 @@ pub enum FileTransportError {
...
 
-        Ok((
-            Self {
-                transport_dir: transport_dir.to_path_buf(),
-            },
-            rx,
-        ))
+        Ok(Self {
+            transport_dir: transport_dir.to_path_buf(),
+            inbound_rx: Some(rx),
```
kaichaosun, line 43: [neutral suggestion] is the Option needed here? It don't see the None is used else where.

kaichaosun (approved):

Issue #111, opened 2026-05-22
Testnet Account Service


jazzz:
## Problem

For MLS to support multiple LocalIdentities, participants inviting new members to the group MUST invite all the LocalIdentities of an Account. 

This means that clients need to be able to retrieve all the associated signing keys for an Account.

## Whats Needed

- Clients can fetch LocalIdentities associated with an AccountAddress
- Accounts can Add new LocalIdentities.
- Data is persisted for at least N days. 

## Suggestions

- Treat Accounts as a Blob.
    - A Signed Bundle that contains { lamport timestamp, number of records, record 1, record 2,... } 
- Service stores one blob for a accountId.
- Signature must match account. 
- Assume AccountAddresses are a Ed25519 Signing key

## Notes

- Focus is on demonstrating λChats ability to support multiple accounts. The cryptography and security of the Account service is not critical at this moment.

kaichaosun:
Dev work finished, the deployment tracked https://github.com/status-im/infra-logos/issues/24

PR #129, opened 2026-06-08
feat: extend the http registry to store account's installations

kaichaosun:
## Summary

Adds an account → device directory so an account can be resolved to its set of installation/device keys. Extends the existing HTTP keypackage registry to store and serve these account-to-device mappings.

## Changes

- **Account directory abstraction** (`account_directory.rs`): introduces `SignedDeviceBundle`, `DeviceSet`, the `AccountAuthority` (account-key signing capability) and `AccountDirectory` (untrusted publish/fetch client) traits, plus the bundle codec and signature verification.
- **Account traits** (`account.rs`, `service_traits.rs`): `LogosAccount` implements `AccountAuthority` alongside `IdentityProvider`; on testnet the account key and device key coincide.
- **Inbox/context integration** (`inbox_v2.rs`, `context.rs`): `publish_device_bundle` upserts this installation's device key into the account's signed bundle (lamport-versioned, re-published idempotently); exposed via `register_account_bundle`.
- **HTTP registry** (`bin/keypackage-registry/*`, `extensions/components/contact_registry*`): new endpoints + store layer to persist and retrieve signed account installation bundles.

Related to https://github.com/logos-messaging/libchat/issues/111


bin/keypackage-registry/src/handlers.rs
```diff
@@ -97,6 +99,96 @@ async fn fetch(
...
+    pub signature: String,
+    /// Unix timestamp (ms) of the last successful upsert.
+    pub updated_at: i64,
+}
+
+/// `POST /v0/account` — upsert a signed device-list bundle for an account.
+///
+/// The server verifies the Ed25519 signature and then stores exactly one blob
+/// per `account_id`, replacing any previous value. Clients should re-publish
+/// whenever they add or rotate LocalIdentities.
```
jazzz: [Pebble] Not a blocker for this work. However this approach is susceptible to replay attacks. Server would need to check a lamport timestamp and only replace the the bundle when it is more recent.

jazzz (approved):
Looks good for the current goals. 


Notes for the Future:

-  The account_id structure will need an update, however it will still contain the verifying key so minimal changes.
- Id recommend keeping the account and keypackage paths as separated as possible, as  the code/functionality will be updated at different times

PR #130, opened 2026-06-08
docs: add target architecture reference

osmaczko:
Design and review discussions have surfaced diverging views of how the chat stack is layered. Add a normative reference for the target architecture so those discussions share one picture.

- Defines the layers — Conversation Core, Client (Generic + Platform), Application — with a responsibilities table and call/event flow.
- States the governing principles: caller-driven sync core, sync results vs async observations, a data-not-behavioural boundary, and easy-edge/powerful-core.
- Documents runtime independence (Generic Client optional; runtime choice at the Platform Client) and the Services/ConversationTypes axis.
- Aligns terminology with the chat specifications.

### DIFF
--- a/docs/target-architecture.md
+++ b/docs/target-architecture.md
@@ -1,0 +1,205 @@
+# Target Architecture
+
+> **Status:** Proposed — for alignment · **Last revised:** 2026-06-08
+
+This document describes the **target architecture** for the chat stack: the role each
+layer plays and the boundaries between them. It is a normative reference for design
+discussions and reviews — a shared picture of the shape we are building toward, not a
+description of any current implementation.
+
+Terminology follows the [chat specifications](https://github.com/logos-messaging/specs)
+(`chatdefs`, `chat-framework`, `privatev1`).
+
+## Vocabulary
+
+| Term | Meaning |
+|---|---|
+| **Content** | Opaque application bytes whose meaning is defined solely by the Application (which may structure them with an envelope such as a ContentFrame). The Core never inspects Content. |
+| **Frame** | A structured protocol message exchanged between Clients. |
+| **Payload** | The serialized binary form of a Frame; opaque to the transport. |
+| **Conversation** | One instance of a chat protocol between participants; owns protocol operations (encryption, reliability, segmentation). |
+| **ConversationType** | A versioned protocol implementation (`PrivateV1`, `GroupV1`, …). Self-contained: defines its own Frame types and storage needs. |
+| **Delivery Service (DS)** | The transport boundary, both directions: accepts outbound Payloads by delivery address and delivers inbound Payloads to the subscribers of an address. |
+| **Service** | An external effect the Core needs but does not implement — delivery, persistence, registration/identity. The Core defines it as a trait; an outer layer supplies the implementation. |
+| **Event** | An asynchronous notification of something the Application could not learn as a return value: an inbound observation (message received, conversation started) or a deferred outcome (delivery acknowledged or timed out).
+| **Client** | The component that manages Conversations and exposes messaging to Applications. |
+| **Application** | Software that integrates a Client to send and receive Content. |
+
+## Layers
+
+Four responsibilities stack from the protocol engine up to the product. Dependencies
+point downward: each layer knows only about the layer beneath it.
+
+```mermaid
+flowchart TB
+    App["<b>Application</b><br/>logos-chat-module + logos-chat-ui"]
+
+    subgraph Client["Client"]
+        direction TB
+        PC["<b>Platform Client</b><br/>concrete Services + ergonomic API"]
+        GC["<b>Generic Client</b> <i>(optional)</i><br/>worker thread + Event queue"]
+    end
+
+    Core["<b>Conversation Core</b><br/>synchronous protocol engine<br/>hosts ConversationTypes · defines Service traits"]
+
+    App -->|calls| PC
+    PC -->|forwards| GC
+    GC -->|locks + calls| Core
+    Core -->|"outcome"| GC
+    GC -. events .-> PC
+    PC -. events .-> App
+
+    style GC stroke-dasharray: 5 5
+```
+
+Solid arrows are synchronous — a call going down and its result coming back on the same
+call; the Conversation Core hands its outcome straight back to its caller.
+Dotted arrows carry Events asynchronously: the Generic Client fills a queue, which the
+Platform Client adapts into the surface the Application consumes. The Application depends
+only on the Platform Client, in both directions — it calls the Platform Client and consumes
+the Event surface the Platform Client exposes. The **Generic Client** and **Platform
+Client** are the two roles of the single *Client* the specs name, shown as the outer box —
+a reusable runtime mechanism, and a concrete binding over it. The Generic Client is
+**optional** (dashed border): a platform without OS threads omits it, and the Platform
+Client drives the Core directly — see [Runtime independence](#runtime-independence).
+
+### Responsibilities
+
+| Layer | Role & owns | Execution model | Must not | Example | Repo |
+|---|---|---|---|---|---|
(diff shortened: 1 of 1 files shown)


docs/target-architecture.md
```diff
@@ -0,0 +1,204 @@
...
+(`chatdefs`, `chat-framework`, `privatev1`).
+
+## Vocabulary
+
+| Term | Meaning |
+|---|---|
+| **Content** | Opaque application bytes whose meaning is defined solely by the Application (which may structure them with an envelope such as a ContentFrame). The Core never inspects Content. |
+| **Frame** | A structured protocol message exchanged between Clients. |
+| **Payload** | The serialized binary form of a Frame; opaque to the transport. |
+| **Conversation** | One instance of a chat protocol between participants; owns protocol operations (encryption, reliability, segmentation). |
```
kaichaosun, line 20: Segmentation belongs to transport layer, not in the chat protocol.
For reliability, it's still blur between chat and delivery, so should be fine putting it here.

docs/target-architecture.md
```diff
@@ -0,0 +1,204 @@
...
+Client drives the Core directly — see [Runtime independence](#runtime-independence).
+
+### Responsibilities
+
+| Layer | Role & owns | Execution model | Must not | Example | Repo |
+|---|---|---|---|---|---|
+| **Conversation Core** | The protocol engine. Hosts pluggable ConversationTypes, defines the Service trait contracts, routes an inbound Payload to the right Conversation, returns a structured outcome. | Strictly synchronous and caller-driven; runs on the caller's thread. | Spawn threads · do background work · call out via callbacks · depend on a runtime or transport. | `Core` + `PrivateV1` / `GroupV1` | `libchat` |
+| **Generic Client** *(optional)* | The native runtime the Core lacks, **batteries included**: an OS worker that drives inbound processing, the Event queue, lifecycle/shutdown — **plus portable default Service implementations** (in-memory, SQLite) so a native app works out of the box. Parameterized over the Service traits. | Owns the worker thread and the Event queue (a buffer it fills). Outbound calls run on the caller's thread. | Call outward into its consumer · depend on an async runtime. | `ChatClient<D, R>` | `libchat` |
```
kaichaosun, line 71: The generic client feels like the api the assemble core and its services together. I would call it `chat-api` if it makes sense. Most use cases will probably consume this API instead of core directly, so making it optional here is not reflect this fact.

And the platform client looks like an adaptor between specific running environment/framework and the chat-api, most chat related business logic will be bridged directly to `chat-api/generic-client`, only the platform specific boilerplate code are different between different platform clients.

docs/target-architecture.md
```diff
@@ -0,0 +1,204 @@
...
+        end
+    end
+
+    GC["<b>Generic Client</b> <i>(optional)</i>"]
+
+    subgraph LogosChat["Platform Client — LogosChat"]
+        direction TB
+        LC_D["LogosDelivery<br/>(through LogosCore)"]
+        LC_P["SQLite"]
+        LC_A["λAccount"]
```
kaichaosun, line 140: Logos Delivery and logos account are from different modules, we should clearly point out the platform client is an bridge for such info.

In my opinion SQLite should lives in chat-api/generic-client as the default persistence layer, though chat-api could provide override methods for a different persistence layer before construct a new chat instance. Since database schemas is heavily couples with each conversations, make such reference storage implementation near to the core will be helpful in reasoning about the code.

docs/target-architecture.md
osmaczko: That's not always the case. In case of PrivateV1 segmentation belongs to the protocol itself. Justification: https://github.com/logos-messaging/specs/blob/master/standards/application/privatev1.md#segmentation

docs/target-architecture.md
```diff
@@ -0,0 +1,205 @@
...
+Client drives the Core directly — see [Runtime independence](#runtime-independence).
+
+### Responsibilities
+
+| Layer | Role & owns | Execution model | Must not | Example | Repo |
+|---|---|---|---|---|---|
+| **Conversation Core** | The protocol engine. Hosts pluggable ConversationTypes, defines the Service trait contracts, routes an inbound Payload to the right Conversation, returns a structured outcome. | Strictly synchronous and caller-driven; runs on the caller's thread. | Spawn threads · do background work · call out via callbacks · depend on a runtime or transport. | `Core` + `PrivateV1` / `GroupV1` | `libchat` |
+| **Generic Client** *(optional)* | The native runtime the Core lacks, **batteries included**: an OS worker that drives inbound processing, the Event queue, lifecycle/shutdown — **plus portable default Service implementations** (in-memory, SQLite) so a native app works out of the box. Parameterized over the Service traits. | Owns the worker thread and the Event queue (a buffer it fills). Outbound calls run on the caller's thread. | Call outward into its consumer · depend on an async runtime. | `ChatClient<D, R>` | `libchat` |
```
jazzz, line 71: From an architectural POV - The distinction between the Generic and Platform clients is hard to differentiate. 

Seems like this could be flattened into a single arch layer, with the construction of clients being an impl concern. 

That is there is only one Client component, thought it may be split over multiple files/structs

docs/target-architecture.md
kaichaosun: It's probably outdated, not reflect current stage? cc @jazzz 
As the DOS, authentication of segments seems resolvable in transport layer, not 100% sure though.

docs/target-architecture.md
```diff
@@ -0,0 +1,205 @@
+# Target Architecture
+
+> **Status:** Proposed — for alignment · **Last revised:** 2026-06-08
+
+This document describes the **target architecture** for the chat stack: the role each
+layer plays and the boundaries between them. It is a normative reference for design
+discussions and reviews — a shared picture of the shape we are building toward, not a
+description of any current implementation.
```
jazzz, line 8: This document attempts to define a canonical form/configuration for independent components. 

Doing so, it blurs across two different separation of concern boundaries. 

`Core -> Client`        `LogosChatClient -> Application`

When it was requested to merge the the Client and Core into the same repo it was done with a promise that they would stay separate logical projects - My fear is that this document attempts to undo that promise.

By defining it as a single "Chat stack" the inherent modularity is discarded. 

I understand that this document provides a high level picture of how messages flow however it is taking a single case and attempting to apply it to all.

My push would be to keep the problems separate.

docs/target-architecture.md
jazzz, line 71: >  Must not: depend on an async runtime.

Perhaps there are additional constraints from your end @osmaczko - I do not see this as a product requirement.

docs/target-architecture.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Governing principles
+
+1. **Simplicity of the core.** The Conversation Core is fully synchronous and
+   caller-driven: no background work, no callbacks out. External effects flow through
+   Services injected as parameters. This keeps the protocol code small and highly
+   reviewable — the basis for trust, audit, and contribution.
+
+2. **Synchronous results, asynchronous observations.** The result of an action the
+   Application takes comes back synchronously, from the call that started it. Everything
```
jazzz, line 83: > The result of an action the
   Application takes comes back synchronously, from the call that started it.

I think this will be hard principal to hold. The process of adding members to a group may take several minutes - Blocking during this would be ill advised

docs/target-architecture.md
```diff
@@ -0,0 +1,205 @@
...
+   goals don't compete because they live in different layers.
+
+## Runtime independence
+
+Because the Core is synchronous and caller-driven (principle 1) and the boundary is
+pull-based data (principle 3), the design commits to a runtime at neither end:
+
+- **The Generic Client is optional.** It is just the native realization of the runtime
+  role, so a platform can skip it and drive the Core itself — for example one without OS
+  threads, which has no use for a worker thread.
```
jazzz, line 108: I think this is fair.

However for now I'd say lets get one client working, and then abstract common functionality once there are multiple.

osmaczko:
Stale, closing.

Issue #109, opened 2026-05-22
Associate Content to an Account, LocalIdentity tuple


jazzz:
## Problem

Groupchats use a signing key which belongs to the localIdentity for identity in MLS. This does not allow multiple LocalIdentities to resolve to single Account.

## Task

- [x] Ensure Applications can see both the Account and LocalIdentity of a message.
- [x] Expand LogosAccounts crate to assist with this look up

PR #137, opened 2026-06-18
feat: receiver recovers both the local identity and the account

kaichaosun:
The changes binds the Account into the MLS leaf credential so a receiver recovers both — verifiably, with no directory round-trip.

- `logos-account`: `credential` module — `encode_credential`  build an account-endorsed credential; `resolve_sender` verifies it against the  leaf device key and returns `MessageSender { account, local_identity }`.
- `conversations`: `get_credential` emits the account-bound credential; `GroupV1`  resolves and attaches the sender; `ConvoOutcome` gains `sender`. GroupV2 surfaces the LocalIdentity (account resolution is a de-mls follow-up).
- `logos-chat`: `Event::MessageReceived` gains `sender`.

Tests: credential codec units + GroupV1 integration assert the verified sender and that distinct identities resolve to distinct accounts. 

Related to https://github.com/logos-messaging/libchat/issues/109

### DIFF
--- a/core/account/src/credential.rs
+++ b/core/account/src/credential.rs
@@ -1,50 +1,71 @@
-//! Account ↔ LocalIdentity binding carried inside the MLS leaf credential.
+//! The MLS leaf credential as a plain Account claim, plus the service that
+//! validates it.
 //!
-//! A group message in MLS is signed by a per-device key — the **LocalIdentity**
-//! (a device/installation). On its own that key says nothing about which
-//! **Account** the device belongs to, so multiple LocalIdentities of one Account
-//! cannot be collapsed back to that Account on the receiving side.
+//! Two identity scopes meet in a group message:
 //!
-//! This module binds the two together *inside the credential itself*, so a
-//! receiver resolves both without any network round-trip or trusted directory:
+//! - **Signer** — the device key. MLS proves the sender controls it; it *is* the
+//!   LocalIdentity. Surfaced as the leaf's signature key.
+//! - **Credential** — a public *claim* `(AccountId, device public key)`. The
+//!   credential content carries only the claimed Account; the device key is the
+//!   leaf signature key MLS hands us.
 //!
-//! - the MLS leaf's signature key is the LocalIdentity device key (MLS already
-//!   proves the sender holds its private half), and
-//! - the credential content carries the Account key plus the Account's signature
-//!   endorsing that exact device key.
-//!
-//! [`resolve_sender`] verifies the endorsement against the device key the leaf
-//! actually signs with, yielding a [`MessageSender`] the application can trust.
-//! This is the inverse of the account → device directory in `libchat`
-//! (`account_directory`): that resolves an Account *to* its devices for invites;
-//! this resolves a device *back to* its Account on receipt.
+//! A claim is not trusted on its own — anyone can staple any account id next to
+//! their own device key. Trust comes from an account service (the account →
+//! device directory): it answers "is this device key actually registered to
+//! that account?". Decoding ([`decode_credential`]) is therefore separate from
+//! validation: core surfaces the raw [`SenderCredential`], and the client
+//! validates it before reporting an identifier to the application.
 
-use crypto::{Ed25519Signature, Ed25519VerifyingKey};
+use crypto::Ed25519VerifyingKey;
 use shared_traits::IdentId;
 use thiserror::Error;
 
 /// Current credential content version. Bump when [`encode_credential`] changes.
 pub const CREDENTIAL_VERSION: u8 = 1;
 
-/// Domain-separation tag prepended to the credential content and folded into the
-/// endorsement message. The account key may sign other things (e.g. the device
-/// bundle in `account_directory`), so binding to this exact purpose stops a
-/// signature obtained elsewhere from being replayed here. The trailing NUL keeps
-/// it from being a prefix of any other domain.
+/// Domain-separation tag prepended to the credential content, so these bytes
+/// can't be confused with any other signed/encoded payload in the system. The
+/// trailing NUL keeps it from being a prefix of any other domain.
 pub const CREDENTIAL_DOMAIN: &[u8] = b"libchat:account-local-identity\0";
 
-/// The verified sender of a group message: which Account, and which device
-/// (LocalIdentity) of that Account it was sent from. Both are hex-encoded
-/// Ed25519 verifying keys.
+/// The raw, *unvalidated* sender of a group message, decoded from the MLS
+/// credential: the claimed Account and the device (LocalIdentity) it was sent
+/// from. Both are hex-encoded Ed25519 verifying keys.
+///
+/// The `local_identity` is trustworthy on its own — MLS verified the message
+/// against that device key. The `account` is only a *claim* until an
+/// [`AccountService`] confirms the device belongs to it.
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub struct MessageSender {
-    /// The Account the sending device belongs to.
+pub struct SenderCredential {
+    /// The Account the sender *claims* to belong to (hex of the account key).
     pub account: IdentId,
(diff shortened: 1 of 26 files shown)


core/conversations/src/conversation/group_v1.rs
```diff
@@ -178,6 +176,26 @@ impl GroupV1Convo {
         &self.convo_id
     }
 
+    /// Resolve the verified [`MessageSender`] for a processed message.
+    ///
+    /// MLS guarantees the message was signed by the sender leaf's key, which is
+    /// the device (LocalIdentity) key. We pair that authoritative key with the
```
jazzz, line 182: [Boulder] This statement is not true. MLS content messages are not signed.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -178,6 +176,26 @@ impl GroupV1Convo {
...
+        let Sender::Member(leaf_index) = processed.sender() else {
+            return None;
+        };
+        // The leaf's signature key is the device key MLS verified the message
+        // against — the trustworthy LocalIdentity, not anything self-asserted in
+        // the credential content.
+        let member = self.mls_group.member_at(*leaf_index)?;
+        let key_bytes: [u8; 32] = member.signature_key.as_slice().try_into().ok()?;
+        let device_key = Ed25519VerifyingKey::from_bytes(&key_bytes).ok()?;
+        logos_account::resolve_sender(processed.credential().serialized_content(), &device_key).ok()
```
jazzz: [Boulder] Do not perform the leaf_key lookup manually. Use the populated Credential field

core/conversations/src/conversation/group_v2.rs
```diff
@@ -460,6 +461,16 @@ impl GroupV2Convo {
                 content: Some(Content {
                     bytes: cm.message.clone(),
                 }),
+                // de-mls carries only the sender's member-id (the LocalIdentity
+                // display string); it does not yet expose an account-bound
+                // credential the way GroupV1 does. Surface the LocalIdentity and
+                // treat it as its own Account (the testnet 1:1 case).
```
jazzz: [!] Demls encodes the credential a s ConversationMessage::sender. You can retrieve a `credential` by deserializing into an `Openmls::Credential`

core/account/src/credential.rs
```diff
@@ -0,0 +1,242 @@
...
+/// endorsement : [u8; 64]  (64 bytes) — account signature over endorsement_message
+/// ```
+pub fn encode_credential(account: &Ed25519VerifyingKey, endorsement: &Ed25519Signature) -> Vec<u8> {
+    let mut out = Vec::with_capacity(CREDENTIAL_DOMAIN.len() + 1 + 32 + 64);
+    out.extend_from_slice(CREDENTIAL_DOMAIN);
+    out.push(CREDENTIAL_VERSION);
+    out.extend_from_slice(account.as_ref());
+    out.extend_from_slice(endorsement.as_ref());
+    out
+}
```
jazzz: I generally like this approach of using a concretely defined payload structure with a prefixed sigil.

core/integration_tests_core/tests/mls_integration.rs
```diff
@@ -57,4 +57,30 @@ fn create_group() {
...
+    // single-device account, so the two resolve to the same key.
+    let raya_sender = harness
+        .saro()
+        .sender_of(&convo_id, M_R1)
+        .expect("Saro should see who sent Raya's message")
+        .clone();
+    assert_eq!(
+        raya_sender.account, raya_sender.local_identity,
+        "single-key testnet account resolves Account == LocalIdentity"
+    );
```
jazzz: [Dust] Consider extending `TestClient::check` to accept a, expected sender. If a test checks that a message arrived it should also check the sender is valid

core/account/src/credential.rs
```diff
@@ -0,0 +1,242 @@
...
+/// Produce the Account's endorsement of `device`, using the Account's signing
+/// capability. `sign` is the account authority's signer (a local key on testnet,
+/// or an external wallet/enclave), so the closure is fallible.
+pub fn endorse_local_identity<E>(
+    account: &Ed25519VerifyingKey,
+    device: &Ed25519VerifyingKey,
+    sign: impl FnOnce(&[u8]) -> Result<Ed25519Signature, E>,
+) -> Result<Ed25519Signature, E> {
+    sign(&endorsement_message(account, device))
+}
```
jazzz: [Pebble] This approach does not meet the requirements. Introducing signed bundled, embedded with each payload makes revocation nearly impossible. 

https://roadmap.logos.co/messaging/furps/application/group_chat.  ->   F6


- Account should support a mechanism for authorizing a LocalSigner.
- Account creates an AccountRecord and submits it to the AccountRegistry. 
    - AccountRecord(AccountData, Signature_over_account_data);
    - AccountData( account_public_key, lamport_timestamp, number of entries, AuthorizedSignerPublicKey,  AuthorizedSignerPublicKey ... );

jazzz (commented):
Love the ambition here, but there are some blockers with the current approach. 

The core issue is that the account crate seems focused on creating credentials. Thats not a terrible approach, however its not quite what we are looking for. This moves message validation to the moment the credential is created, which causes issues with the authentication model. 

We need lazy evaluation at the Client level, in order to support revocation.

core/account/src/credential.rs
kaichaosun: Yeah, good to know that, it was proposed before I read your message to use account service as the verification check for device -> account.
I just updated to the new approach.

core/conversations/src/conversation/group_v1.rs
kaichaosun: `ProcessedMessage` doesn't expose the signing key, `credential()` returns a Credential, which is just `{ credential_type, serialized_credential_content }` (our account claim); the device key isn't on it. 
OpenMLS builds a CredentialWithKey during verification but doesn't surface it on the processed message. The leaf's signature_key is only available via the ratchet tree (member_at).

If we instead embedded the device key in the credential content and read it back, it'd be self-asserted. MLS doesn't enforce that content matches the leaf's real signature_key, so a member could put a different key there.

Is there a populated field you had in mind that I'm missing?

core/conversations/src/conversation/group_v1.rs
kaichaosun: It looks processed messages are checked with signing methods,
https://github.com/openmls/openmls/blob/main/openmls/src/group/mls_group/processing.rs#L591

core/conversations/src/conversation/group_v2.rs
kaichaosun: de-mls verifies each message against its own generated leaf signing key, which is not libchat's device key. 
So unlike group-v1 where the MLS leaf key is the device key, groupv2 has no link between the key it verifies and libchat's account/device model. That's the whole gap.

I updated a bit, but looks like there are more work needed, and I'm not sure if I'm in the correct direction. To split the concern, let's keep the scope of this PR on group-v1. 

https://github.com/vacp2p/de-mls/pull/113

jazzz: >  groupv2 has no link between the key it verifies and libchat's account/device model. That's the whole gap.

GroupV2 will use the same keypackages and MLS config as GroupV1. Its some work to migrate the code, but that is the integration point. Crendentials will contain an LocalSignerId, and an AccountId. 

> de-mls verifies each message against its own generated leaf signing key

For completeness this is technically not true. What is true is that openmls associates a received content message with a given credential. If the wiring from openmls::ProcessedMessage to ConvoOutcome may be missing, happy to complete this if that would help.  


> Let's keep the scope of this PR on group-v1.

Agreed. Or go even higher, and implement strictly at the Core/Client interface and the wiring can be completed later.

kaichaosun:
Replaced by https://github.com/logos-messaging/libchat/pull/142

Issue #114, opened 2026-05-22
Implement 1:1 Chats using Groups


jazzz:
## Problem

Currently 1:1 chats do not support HNDL protections via post-quantum encryption. Updating them would be be costly for limited benefit a this point. 

## Whats Needed

- [x] Create PrivateV2Convo which is based on groups using InboxV2
- [x] Limit membership to two participants
- [x] Update create_private_convo to initialize PrivateV2Convo

## Notes

- Exploring PQ 1:1 chats should be explored at a later point due to privacy implications  
   - Currently lacking PQ:HNDL is a bigger issue.
- Verification needed: It should be possible to use base openMLS (Implemented In GroupV1) in a two person group. De-MLS adds consistency for multi-party state updates, however the 2 participant case is less complex.

jazzz:
Conversation is initialized on `create_direct_convo` instead of `create_private_convo` as the later path, breaks all test instances.

This is separated into followup work #144

PR #139, opened 2026-06-18
Add 1:1 Chats using Groups

jazzz:
This PR partially addresses #114.

It adds a new ConversationType which wraps GroupV1Convo, however only exposes the `Convo` Trait. This limits membership to 2 people enforced by the constructor.

## Notes
- Rather than naming the ConvoType `PrivateV2Convo`, I chose `DirectV1Convo`. 'Private' Implies that 'Group' is not private. Direct is a better name as it describes the unique feature, that its direct between two participants. This borrows from the "DirectMessage" concept which would be implemented using this convoType
- This PR does not enable DirectV1 as the default convo type. 
   - The method signatures are different as the new type does not accept `Introduction`'s ; Will handle in follow-up PR.
- `ConversationTypeOwned::Direct` is added to track the ConvoClass. No Further classes are expected, and may be removed in the future.
- Calling `add_member` on `ConversationTypeOwned::Direct` throws a runtime error. This will need to be addressed in the future by extending ConversationId to be type aware, which provides type safety

### DIFF
--- a/core/conversations/src/conversation/direct_v1.rs
+++ b/core/conversations/src/conversation/direct_v1.rs
@@ -21,7 +21,6 @@ impl DirectV1Convo {
     // have multiple Installations.
     pub fn new<S: ExternalServices>(
         cx: &mut ServiceContext<S>,
-        // Constructor must accept multiple
         members: &[IdentIdRef],
     ) -> Result<Self, ChatError> {
         let mut inner_group = DelegateGroup::new(cx)?;
--- a/core/conversations/src/conversation/direct_v1.rs
+++ b/core/conversations/src/conversation/direct_v1.rs
@@ -10,11 +10,11 @@ use crate::{
 type DelegateGroup = GroupV1Convo;
 
 #[derive(Debug)]
-pub struct PrivateV2Convo {
+pub struct DirectV1Convo {
     inner_group: DelegateGroup,
 }
 
-impl PrivateV2Convo {
+impl DirectV1Convo {
     pub fn new<S: ExternalServices>(
         cx: &mut ServiceContext<S>,
         participant: IdentIdRef,
@@ -25,13 +25,13 @@ impl PrivateV2Convo {
     }
 }
 
-impl Identified for PrivateV2Convo {
+impl Identified for DirectV1Convo {
     fn id(&self) -> ConversationIdRef<'_> {
         self.inner_group.id()
     }
 }
 
-impl<S> Convo<S> for PrivateV2Convo
+impl<S> Convo<S> for DirectV1Convo
 where
     S: ExternalServices,
 {
--- a/core/conversations/src/conversation/direct_v1.rs
+++ b/core/conversations/src/conversation/direct_v1.rs
@@ -9,18 +9,23 @@ use crate::{
 
 type DelegateGroup = GroupV1Convo;
 
+/// A Conversation between two participants.
 #[derive(Debug)]
 pub struct DirectV1Convo {
     inner_group: DelegateGroup,
 }
 
 impl DirectV1Convo {
+    // Constructor must accept multiple IdentId's
+    // While the conversation is limited to 2 participants, each participants may
+    // have multiple Installations.
     pub fn new<S: ExternalServices>(
         cx: &mut ServiceContext<S>,
-        participant: IdentIdRef,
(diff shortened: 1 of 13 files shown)


jazzz:
PR is based on a feature branch to keep reviews small. This PR must be merged into `main` after #138  lands

osmaczko (approved):
Clean 👍

core/conversations/src/conversation/direct_v1.rs
```diff
@@ -0,0 +1,62 @@
...
+    ChatError, ExternalServices,
+    conversation::{ConversationIdRef, Convo, GroupConvo, GroupV1Convo, Identified},
+    service_context::ServiceContext,
+};
+
+type DelegateGroup = GroupV1Convo;
+
+/// A Conversation between two participants.
+#[derive(Debug)]
+pub struct DirectV1Convo {
```
kaichaosun, line 14: [suggestion] The version in file and struct make code hard to reason about in the long term with different versions cross referenced in one or many files. 
If it's hard to to find a proper name, we should probably just use spec id/number (the spec doesn't need to be perfect, it can be early and rough, or maintained in our codebase), since the spec will include the necessary context, which also helps communication among developers about the indention of specific convo protocol.

core/conversations/src/conversation/direct_v1.rs
```diff
@@ -0,0 +1,62 @@
...
+    inner_group: DelegateGroup,
+}
+
+impl DirectV1Convo {
+    // Constructor must accept multiple IdentId's
+    // While the conversation is limited to 2 participants, each participants may
+    // have multiple Installations.
+    pub fn new<S: ExternalServices>(
+        cx: &mut ServiceContext<S>,
+        // Constructor must accept multiple
```
kaichaosun:
```suggestion
```

core/conversations/src/conversation/direct_v1.rs
```diff
@@ -0,0 +1,62 @@
...
+        members: &[IdentIdRef],
```
kaichaosun, line 24: [request change] What's the `IdentIdRef` means here, is it devices or accounts? We should avoid using general terms like `iden id`.

Also `members` is confusing, it should just be the opposing side of a direct chat. Multiple devices support needs a better architecture, and the current codebase is not providing it yet.

core/conversations/src/conversation/direct_v1.rs
jazzz: As per [ConversationTypes](https://github.com/logos-co/logos-lips/blob/master/docs/messaging/application/raw/conversationtypes.md#protocol-naming), there are no versions. Each conversationType is a standalone type.

> If it's hard to to find a proper name,

> we should probably just use spec id/number

This isn't a bad idea, and easy to implement later. Using spec ID's such as `Convo143` I *think* would require a change to LIPS process to ensure stable identifiers over spec revisions. But I'll be honest I'm not sure how this addresses ".. hard to reason about in the long term with different versions cross referenced in one or many files."

core/conversations/src/conversation/direct_v1.rs
jazzz: Agreed  whole heartedly on `IdentIdRef`- This PR doesn't introduce `IdentIdRef` so changing it would be a follow up PR as it touches many files.

> it should just be the opposing side of a direct chat.

Yes. 
```
// Constructor must accept multiple IdentId's
// While the conversation is limited to 2 participants, each participant may
// have multiple Installations."
```
DirectConvo is a "GroupChat" with many members. They happen to correspond to 2 participants.

jazzz: > Also members is confusing,

Suggestions Welcome

jazzz: > Multiple devices support needs a better architecture, and the current codebase is not providing it yet.

The core is currently agnostic to all concepts of Multidevice support - That is a client concern.  
It only concerns it self with IdentityProviders and Id's. So there should be no blockers on infra.

> What's the IdentIdRef means here, is it devices or accounts? 
Its a member of a conversation. Core SHOULD only operate on "devices". An account is a group of Devices.

This Multi-device support infra in client is part of #111 and #109

PR #140, opened 2026-06-19
Add client path for DirectConvo

jazzz:
This PR updates client with the ability to create DirectConversations as per #139. This provides an easy integration point for future client work. 

## Notes
- Deprecating `create_conversation` is deferred as waiting on #111
- Adds place holder look up function for Accounts -> LocalSigners. This is premature, however documents the expected flow. It can be completely replaced when needed.

 !TODO: Create issue to remove `create_conversation` and update tests/apps


crates/client/src/client.rs
```diff
@@ -158,8 +160,24 @@ where
...
+        self.core
+            .lock()
+            .create_direct_convo(&signer_refs)
+            .map_err(Into::into)
+    }
+
     /// Parse intro bundle bytes and initiate a private conversation. Outbound
     /// envelopes are published by the core. Returns this side's conversation ID.
+    ///
+    /// This function will be deprecated in the future. Use `create_direct_conversation`
```
osmaczko, line 180: Can't we deprecate it now?
```suggestion
    #[deprecated(note="use `create_direct_conversation` instead")]
```

osmaczko (approved):

crates/client/src/client.rs
jazzz: It introduces warnings across the apps and tests. Without the infrastructure for GroupsChats, there's nothing to migrate the tests too.

PR #141, opened 2026-06-19
update de-mls setup

seemenkina:


core/conversations/src/conversation/group_v2.rs
```diff
@@ -184,12 +180,23 @@ impl GroupV2Convo {
...
         service_ctx: &mut ServiceContext<S>,
     ) -> Result<Self, ChatError> {
-        let setup = DemlsSetup::new(service_ctx.mls_identity.display_name())?;
         let convo_id = rand_string(5);
-        let conversation = Conversation::create(&convo_id, setup.deps())?;
+        let member = member_id(service_ctx);
+        let conversation = Conversation::create(
+            &convo_id,
+            &service_ctx.mls_provider,
+            service_ctx.mls_identity.get_credential(),
```
jazzz, line 188: [!] I like this. I can cleanly change the behavior of the group by passing in different components/configurations

core/conversations/src/conversation/group_v2.rs
```diff
@@ -97,73 +78,88 @@ impl IdentityProvider for NamespacedIdentity<'_> {
...
+/// (storing its private keys there so the matching welcome opens), and return
+/// the serialized public key package.
+fn mint_key_package<S: ExternalServices>(
+    service_ctx: &ServiceContext<S>,
+) -> Result<Vec<u8>, ChatError> {
+    let capabilities = Capabilities::builder()
+        .ciphersuites(vec![CIPHER_SUITE])
+        .extensions(vec![ExtensionType::ApplicationId])
+        .build();
+    let bundle = KeyPackage::builder()
```
jazzz, line 143: [?] Is there a driver for GroupV2 to create its own keyPackages? Or could I replace this in a followup PR?

InboxV2 handles the creation and publication of KeyPackages.
ref: https://github.com/logos-messaging/libchat/blob/main/core/conversations/src/inbox_v2.rs#L202-L221

--

If there is no need; See any blockers with me merging `new_pending` and `accept_welcome` into one function `new_from_welcome`?

core/conversations/src/conversation/group_v2.rs
```diff
@@ -357,8 +382,8 @@ where
         members: &[IdentIdRef],
     ) -> Result<(), ChatError> {
         // Record who WE invited before touching the conversation: after_op
-        // forwards a welcome only to joiners in pending_invites (member-id
-        // bytes == account name bytes for LocalDemlsMember).
+        // forwards a welcome only to joiners in pending_invites (the de-mls
+        // member-id is the invitee's account id bytes).
```
jazzz: Were still working on our Account/Delegate language, suggest not explicitly saying 'account' to avoid confusion.

```suggestion
        // member-id is the invitee's id bytes).
```

core/conversations/src/inbox_v2/identity.rs
```diff
@@ -16,6 +16,7 @@ use crate::IdentityProvider;
 /// This type stops OpenMLS internal from leaking outside of the crate.
 /// Developers provider a simple IdentitityProvider, and Signer and Credential generation
 /// is provided
+#[derive(Clone)]
```
jazzz: [Pebble] Cloning the Provider may create duplicate storage instances leading to split state. Will eventually need to be a passed reference

jazzz (approved):
Overall this is improved. Like the direction

core/conversations/src/conversation/group_v2.rs
seemenkina: Sure

core/conversations/src/inbox_v2/identity.rs
seemenkina: It's actually was from previous idea, so now it isn't necessary rn, I'll remove it

core/conversations/src/conversation/group_v2.rs
seemenkina: > Is there a driver for GroupV2 to create its own keyPackages?

de-mls has no KP driver anymore — minting is the integrator's job, so move it whenever suitable; my mint_key_package is just a small example. 

>  See any blockers with me merging new_pending and accept_welcome into one function new_from_welcome?

I think their also a little bit outdated functions split up. new_pending is just KP registration and accept_welcome is just Conversation::join + the local group setup (delivery service, etc.). So once InboxV2 owns KP registration, new_pending can be removed and accept_welcome becomes the single new_from_welcome.

jazzz (approved):
Good updates

PR #142, opened 2026-06-21
Verify sender account ↔ device mapping before delivery

kaichaosun:
## Problem

Received messages carry a `DelegateCredential` binding a delegate device key (LocalIdentity) to an optional account address. The account a message claimed to act for was taken at face value, so any device could assert any account.

## Changes

- At the client receive path — `should_deliver`, run from `convo_events` / `inbox_events`, covering both GroupV1 and GroupV2 — check each message's credential against the account → device directory (`AccountDirectory`, our account store).
- Drop a message when its account → device mapping is wrong or can't be confirmed; deliver when it is verified or when no account is claimed. Dropped messages never reach the application.
- Add `Core::account_directory()` to reach the directory from the event layer, and `DelegateCredential` accessors (`delegate_id`, `account_addr`).
- `direct_v1_integration` now publishes a signed device bundle per account so its associated delegates verify end to end; unit tests cover every branch.

### Delivery policy

| Sender credential | Result |
|---|---|
| None (PrivateV1 placeholder) or unassociated — no account claimed | Deliver |
| Associated; account's published device set includes the device | Deliver (verified) |
| Associated; device not in the set / no published set / address not a key / malformed / lookup failed | Drop |

Only the drop paths log (one `warn!` each); verified delivery is silent.

Related to #109. Builds on #143.

### DIFF
--- a/core/conversations/src/core.rs
+++ b/core/conversations/src/core.rs
@@ -153,6 +153,14 @@ impl<'a, S: ExternalServices + 'static> Core<S> {
         &self.services.store
     }
 
+    /// The account → device directory (our account store). Used to verify that a
+    /// received message's claimed account actually endorses the sending device
+    /// before the message is surfaced. Exposed as `RegistrationService`, whose
+    /// `AccountDirectory` supertrait provides `fetch`.
+    pub fn account_directory(&self) -> &S::RS {
+        &self.services.registry
+    }
+
     pub fn identity(&self) -> &Identity {
         &self.services.identity
     }
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -3,9 +3,11 @@ use std::thread::{self, JoinHandle};
 
 use components::{EphemeralRegistry, ThreadedWakeupService, WakeupEvent};
 use crossbeam_channel::{Receiver, Sender, select};
+use crypto::Ed25519VerifyingKey;
 use libchat::{
-    ChatError, ChatStorage, ConversationId, ConvoOutcome, Core, DeliveryService, IdentId,
-    IdentIdRef, InboxOutcome, Introduction, PayloadOutcome, RegistrationService, StorageConfig,
+    AccountDirectory, ChatError, ChatStorage, ConversationId, ConvoOutcome, Core, DeliveryService,
+    IdentId, IdentIdRef, InboxOutcome, Introduction, PayloadOutcome, RegistrationService,
+    StorageConfig,
 };
 use parking_lot::Mutex;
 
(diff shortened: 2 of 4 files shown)


core/conversations/src/conversation/group_v1.rs
```diff
@@ -175,6 +175,27 @@ impl GroupV1Convo {
...
+    fn decode_sender(&self, processed: &ProcessedMessage) -> Option<MessageSender> {
+        let Sender::Member(leaf_index) = processed.sender() else {
+            return None;
+        };
+        let member = self.mls_group.member_at(*leaf_index)?;
+        Some(MessageSender {
+            account: IdentId::new(
+                String::from_utf8_lossy(member.credential.serialized_content()).into_owned(),
+            ),
+            local_identity: IdentId::new(hex::encode(member.signature_key.as_slice())),
```
kaichaosun: I use group member signature key as device/local identity, because it's trustless.
It maybe changed when new infra for devices and accounts comes in.

crates/client/src/event.rs
```diff
@@ -23,6 +23,7 @@ pub enum Event {
     MessageReceived {
         convo_id: Arc<str>,
         content: Vec<u8>,
+        sender: Option<MessageSender>,
```
osmaczko: In theory, `MessageSender` should never be optional, although I understand that for 1-to-1 conversations it is redundant. Would it be possible to make it always available to simplify the app’s logic?

osmaczko (approved):

core/conversations/src/core.rs
```diff
@@ -153,6 +153,14 @@ impl<'a, S: ExternalServices + 'static> Core<S> {
...
     }
 
+    /// The account → device directory (our account store). Used to verify that a
+    /// received message's claimed account actually endorses the sending device
+    /// before the message is surfaced. Exposed as `RegistrationService`, whose
+    /// `AccountDirectory` supertrait provides `fetch`.
+    pub fn account_directory(&self) -> &S::RS {
+        &self.services.registry
+    }
+
```
jazzz, line 163: If I'm reading this correctly. This is needed because the core becomes the owner of the RegistrationService.

[Sand] The ownership of the registry will to shift to the client in the future, such that it can be passed into the service context, while maintaining the ability to manage it directly.

core/conversations/src/core.rs
```diff
@@ -153,6 +153,14 @@ impl<'a, S: ExternalServices + 'static> Core<S> {
         &self.services.store
     }
 
+    /// The account → device directory (our account store). Used to verify that a
+    /// received message's claimed account actually endorses the sending device
+    /// before the message is surfaced. Exposed as `RegistrationService`, whose
+    /// `AccountDirectory` supertrait provides `fetch`.
+    pub fn account_directory(&self) -> &S::RS {
+        &self.services.registry
+    }
```
jazzz, line 162: This combines two different concepts Keypackage registration and Account management.

[Sand] Managing accounts should exclusively be managed via the LogosAccount crate. Eventually the AccountDirectory trait will need to be removed and owned exclusively by the Client

jazzz (approved):
Looks good. 

Theres some future changes needed, but no blockers.

As LogosAccount gets built out we'll need to look at caching values. But thats a future concern

core/conversations/src/core.rs
kaichaosun: Yeah, will move the registry soon.

PR #143, opened 2026-06-22
Add Delegate Signer and wire into Client

jazzz:
## Problem

Chat core uses λ-Accounts as its IdentityProvider which is the incorrect path for MultiDevice support. Instead the clients should use a local signer, with an associated Account. 

## Changes

- Introduces `DelegateSigner` — an Ed25519 keypair identity that can be standalone or associated with an account address. Replaces the `TestLogosAccount` placeholder in `ChatClient` constructors.
- Adds `DelegateCredential`, a TLV-serialized credential binding a delegate key to an optional account address, with `TryFrom<Vec<u8>>` and `TryFrom<IdentId>` round-trips.
- Propagates the sender's credential through the message pipeline: `group_v1` extracts it from the MLS BasicCredential, `group_v2` now uses the full `IdentId` (not the truncated display name) so the receiver can decode it. `privatev1` sends an empty credential as a placeholder.
- Decodes the sender credential on every received message (both `convo_events` and `inbox_events`), logging it at `debug` level. The decode point is marked as the integration point for future authorization checks.
- Adds `ChatClient::new_ephemeral` for tests and integrations that supply their own `DelegateSigner` and `RegistrationService`. Returns `Result` to propagate construction failures from fallible registries.
- Adds `direct_v1_integration` test: two associated delegates create a direct conversation end-to-end.

### DIFF
--- a/bin/chat-cli/src/app.rs
+++ b/bin/chat-cli/src/app.rs
@@ -5,7 +5,7 @@ use std::path::{Path, PathBuf};
 use anyhow::Result;
 use arboard::Clipboard;
 use crossbeam_channel::Receiver;
-use logos_chat::{ChatClient, DeliveryService, EphemeralRegistry, Event, RegistrationService};
+use logos_chat::{ChatClient, EphemeralRegistry, Event, RegistrationService, Transport};
 use serde::{Deserialize, Serialize};
 
 use crate::utils::now;
@@ -41,7 +41,7 @@ pub struct AppState {
     pub active_chat: Option<String>,
 }
 
-pub struct ChatApp<T: DeliveryService, R: RegistrationService = EphemeralRegistry> {
+pub struct ChatApp<T: Transport, R: RegistrationService = EphemeralRegistry> {
     pub client: ChatClient<T, R>,
     events: Receiver<Event>,
     pub state: AppState,
@@ -55,7 +55,7 @@ pub struct ChatApp<T: DeliveryService, R: RegistrationService = EphemeralRegistr
 
 impl<T, R> ChatApp<T, R>
 where
-    T: DeliveryService + Send + 'static,
+    T: Transport,
     R: RegistrationService + Send + 'static,
 {
     pub fn new(
--- a/core/conversations/src/outcomes.rs
+++ b/core/conversations/src/outcomes.rs
@@ -13,6 +13,7 @@ use crate::conversation::ConversationId;
 #[derive(Debug, Clone)]
 pub struct Content {
     pub bytes: Vec<u8>,
+    pub encoded_credential: Vec<u8>,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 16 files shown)


crates/client/src/delegate.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+/// A credential issued to a delegate key, optionally bound to an account address.
+///
+/// Serialized as a TLV byte sequence prefixed with magic bytes `0x23 0x23`.
+/// A credential without an `account_addr` is *unassociated* — it identifies the
+/// delegate key but has not yet been linked to an account.
+#[derive(Debug)]
+pub struct DelegateCredential {
+    delegate_id: Ed25519VerifyingKey,
```
kaichaosun, line 70: [Question] What if a malicious member use the same credential from an honest user?
The credential is not self certified or authencated by centralized entities like CA.

core/conversations/src/outcomes.rs
```diff
@@ -13,6 +13,9 @@ use crate::conversation::ConversationId;
 #[derive(Debug, Clone)]
 pub struct Content {
     pub bytes: Vec<u8>,
+    /// Hex-encoded [`DelegateCredential`] of the sender, if present in the message.
+    /// Empty when the sender did not attach a credential.
+    pub encoded_credential: Vec<u8>,
```
kaichaosun, line 18: [suggestion] It's better to make it optional, so non-mls convos don't need to present this field.

crates/client/src/delegate.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    pub fn associated(delegate: &Ed25519VerifyingKey, account: &str) -> Self {
+        Self {
+            delegate_id: delegate.clone(),
+            account_addr: Some(account.to_string()),
+        }
+    }
+
+    pub fn serialize(self) -> Vec<u8> {
```
kaichaosun, line 92: [request change] need unit test for serialize and deserialize.

core/conversations/src/outcomes.rs
jazzz: All content needs some sender information, regardless of how it is built. DelegateCredential is not based on MLS its based on our approach to multi-device support.


Additionally as the library will only contain MLS based conversations, seems appropriate to follow this path until it is replaced

crates/client/src/delegate.rs
jazzz: Great Question.  

As you brought up in a previous PR. Messages are signed by DelegateSigner. If the Signing Identity does not match the credential that has been provided the message would be considered invalid.

The credential is validated at the edge, rather than by a CA. This area will receive an update and full plan after V0.2

bin/chat-cli/src/app.rs
```diff
@@ -55,7 +55,7 @@ pub struct ChatApp<T: DeliveryService, R: RegistrationService = EphemeralRegistr
 
 impl<T, R> ChatApp<T, R>
 where
-    T: DeliveryService + Send + 'static,
+    T: Transport,
```
osmaczko, line 58: 👍

osmaczko (approved):

PR #145, opened 2026-06-22
ChatClient migration

jazzz:
This PR continue preparing Cli-Client for group chats. Specifically It introduces a Builder pattern which allows clients to be constructed with both testing and production components. 

Relating to De-MLS, this PR:
- Allow Clients to accept any IdentityProvider to support DelegateSigners.
- Removes reliance on legacy settings. 
- Chooses a default implementation for each service.

## Summary

- Introduces `ChatClientBuilder`, a typestate builder for constructing `ChatClient` that makes all four configuration slots (`identity`, `transport`, `registration`, `storage`) explicit at the type level, with sensible defaults for any unset field
- Promotes `ChatClient` from a 2-param `<T, R>` type to a fully generic `<I, T, R, S>` type, making the identity provider and storage backend first-class type parameters; removes the old `open` / `open_with_registry` constructors in favour of the builder
- Adds `storage_config()` on the builder as a convenience that accepts a `StorageConfig` and constructs `ChatStorage` inline, so callers don't need to import `ChatStorage` directly
- Fixes re-entrant message filtering in `GroupV1Convo`: echoed outbound messages (including MLS Commits from `add_member`) are now suppressed via a bounded hash cache (`OUTBOUND_HASH_CACHE_SIZE = 25`) to prevent the sender from re-processing its own messages on loopback transports
- Re-exports `ChatStorage`, `ChatStore`, and `IdentityProvider` from `logos-chat` so downstream crates don't need to reach into internal dependencies
- Updates `chat-cli` (`app.rs`, `ui.rs`, `main.rs`) and the `message-exchange` example to use the new 4-param `ChatApp<I, T, R, S>` and builder API

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3605,6 +3605,7 @@ dependencies = [
  "logos-account",
  "parking_lot",
  "shared-traits",
+ "storage",
  "tempfile",
  "thiserror",
  "tracing",
--- a/bin/chat-cli/src/app.rs
+++ b/bin/chat-cli/src/app.rs
@@ -5,7 +5,7 @@ use std::path::{Path, PathBuf};
 use anyhow::Result;
 use arboard::Clipboard;
 use crossbeam_channel::Receiver;
-use logos_chat::{ChatClient, EphemeralRegistry, Event, RegistrationService, Transport};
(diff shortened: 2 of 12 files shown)


osmaczko (approved):
I like the builder pattern 👷‍♂️

PR #146, opened 2026-06-23
feat: bubble up message sender to applications

kaichaosun:
Adds `sender: Option<MessageSender>` to `Event::MessageReceived` so applications can see the Account and LocalIdentity (device) behind each message.

The sender is decoded from the message credential and checked against the account → device directory:

| sender                                       | Meaning                                |
|----------------------------------------------|----------------------------------------|
 | None                                         | No credential (e.g. PrivateV1 1:1)     |
 | Some { account: None, local_identity }       | Sender claims no account — device only |
 | Some { account: Some(addr), local_identity } | Account verified against the directory |

Spoofed or unconfirmable senders are already dropped upstream, so a present account is always verified — no status flag needed.
 
Related to #109.

### DIFF
--- a/crates/client/src/event.rs
+++ b/crates/client/src/event.rs
@@ -8,7 +8,20 @@
 
 use std::sync::Arc;
 
-use libchat::ConversationClass;
+use libchat::{ConversationClass, IdentId};
+
+/// The sender of a received message, recovered from its credential.
+///
+/// `account` is present only when the sender associated an account *and* the
+/// account → device directory confirmed this device belongs to it — spoofed or
+/// unconfirmable claims never reach the application, so a `Some` account is
+/// always verified. `local_identity` is the sending device (delegate key),
+/// hex-encoded.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MessageSender {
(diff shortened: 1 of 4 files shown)


crates/client/src/event.rs
```diff
@@ -8,7 +8,20 @@
...
+///
+/// `account` is present only when the sender associated an account *and* the
+/// account → device directory confirmed this device belongs to it — spoofed or
+/// unconfirmable claims never reach the application, so a `Some` account is
+/// always verified. `local_identity` is the sending device (delegate key),
+/// hex-encoded.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MessageSender {
+    pub account: Option<IdentId>,
+    pub local_identity: IdentId,
```
jazzz, line 23: [Sand] Support for PrivateV1 is being dropped. I wouldn't bother with the Option<_>. I think its fair to say all messages being returned to an application require an explicit sender.

If there are any messages without a "Sender" they ought to be returned as an SystemEvent, or some future Anonymous tag

jazzz (approved):

PR #148, opened 2026-06-24
Componentize logos delivery

jazzz:
## Problem

The embedded version of LogosDelivery is only available via chat-cli. 

This:
1. Makes it difficult to use the embedded version of logos-delivery in other projects.
2. Complicates chat-cli with transport level concepts and complicated build steps.

## Solution

- Move the `chat-cli/logos-delivery` transport to `components/EmbeddedP2pDeliveryService`
- Remove requirement for build.rs in consumers

### Build Updates

- `extensions/components/build.rs` — single source of truth: finds the lib (env override → nix auto-build), stamps an absolute install name (macOS install_name_tool, Linux patchelf --set-soname), emits the propagating link directives.
- `bin/chat-cli/build.rs` — deleted.
- `extensions/components/src/lib.rs` — export gated on logos_delivery to stay consistent with the module gate in delivery.rs, so the type is present exactly when the lib is available.

### Consuming

`EmbeddedP2pDeliveryService` is available to applications via cargo import.  

`components = { ..., features = ["embedded_p2p_delivery"]}`


bin/chat-cli/Cargo.toml
```diff
@@ -9,9 +9,11 @@ path = "src/main.rs"
 
 [dependencies]
 # Workspace dependencies (sorted)
+components = { workspace = true , features = ["embedded_p2p_delivery"]}
 crossbeam-channel = { workspace = true }
 logos-chat = { workspace = true }
 
+
```
kaichaosun:
```suggestion
```

bin/chat-cli/src/main.rs
```diff
@@ -13,13 +13,33 @@ use logos_chat::{
     RegistrationService, StorageConfig, Transport,
 };
 
+use components::{EmbeddedP2pDeliveryService, P2pConfig};
+
+#[derive(Debug)]
+struct P2pTransport(EmbeddedP2pDeliveryService);
```
kaichaosun, line 19: [question] why wrap the `EmbeddedP2pDeliveryService` into `P2pTransport`, not use `EmbeddedP2pDeliveryService` and implement the traits directly on it?

kaichaosun (approved):

bin/chat-cli/src/main.rs
jazzz: The issue here is who is the owner of the Transport trait?

Transport trait is the result of the ChatClient being based on threaded model - it mandates that a DS uses a `crossbeam` queue to pass events across threads. 

Given this requirement is unique to ChatClient, it doesn't quite make sense to force all DS's operate in this manner.

The approach is to use the newtype pattern to implement Transport where its needed. In the future this can be cleaned further, by building a blanket impl that runs the DS in a thread and captures output into a queue, so this does not require to be implemented.

osmaczko (approved):
Nice cleanup 👍

PR #149, opened 2026-06-24
CI Speed-up

jazzz:


jazzz:
ref: https://github.com/logos-messaging/libchat/pull/150 touches on the same part. Merge that first, to minimize conflicts

PR #150, opened 2026-06-24
feat: update nixpkgs locked version to avoid override nixpkgs in ci.

kaichaosun:
Update nixpkgs pinned in flake.lock and remove override personal branch.


jazzz (approved):
🙏

PR #151, opened 2026-06-25
feat: default value for logos chat client

### LINKED ISSUE #126: Create a LogosChatClient, which is default prefconfigured with Logos Specific services
Context: https://github.com/logos-messaging/libchat/pull/124/changes#r3344852849

Need to deploy the contact store service first.

kaichaosun:
Adds an opinionated `LogosChatClient` preconfigured with the Logos stack.

Changes:
- `LogosChatClient`  with `open(transport, db_path, db_key, registry_url)`. 
  `db_path`/`db_key` are caller-supplied; 
  `registry_url` overrides the preconfigured endpoint  (placeholder until deployed).
- chat-cli builds via `LogosChatClient::open`, so it always uses the registry;
  `--registry-url` is now an endpoint override. 
  Removed the dead `run_logos_delivery`.

Fix https://github.com/logos-messaging/libchat/issues/126

### DIFF
--- a/crates/client/src/logos.rs
+++ b/crates/client/src/logos.rs
@@ -1,0 +1,61 @@
+//! The opinionated Logos client.
+//!
+//! [`ChatClientBuilder`] is generic and can only default the zero-config
+//! components (random identity, ephemeral registry, in-memory storage) — it has
+//! no way to know a registry endpoint or a database path, so its defaults are
+//! the test-grade ones. `LogosChatClient` is the layer that *does* commit to a
+//! stack: a delegate identity, the HTTP keypackage + account registry, and
+//! encrypted on-disk storage. It exists so independently built clients share the
+//! same production services instead of each re-deriving them.
+//!
+//! Only the transport is left to the caller: it carries native dependencies and
+//! environment-specific configuration that belong to the binary, not here.
+
+use crossbeam_channel::Receiver;
+use libchat::{ChatStorage, StorageConfig};
+
+use crate::ChatClientBuilder;
+use crate::client::{ChatClient, Transport};
+use crate::delegate::DelegateSigner;
+use crate::errors::ClientError;
+use crate::event::Event;
+use components::HttpRegistry;
+
+// The endpoint for account and keypackage registration service.
+const REGISTRY_ENDPOINT: &str = "https://devnet.chat-kc.logos.co";
+
+/// A [`ChatClient`] wired to the Logos service stack: a [`DelegateSigner`]
+/// identity, the HTTP keypackage + account registry ([`HttpRegistry`], which is
+/// both the keypackage store and the account → device directory), and encrypted
+/// [`ChatStorage`]. Only the transport `T` is supplied by the caller.
+pub type LogosChatClient<T> = ChatClient<DelegateSigner, T, HttpRegistry, ChatStorage>;
+
+impl<T> LogosChatClient<T>
+where
+    T: Transport + Send + 'static,
+{
+    /// Open a client on the Logos stack over `transport`, persisting to the
+    /// encrypted database at `db_path` unlocked with `db_key`. When `registry_url`
+    /// is `Some`, it overrides the preconfigured registry endpoint (e.g. a local
+    /// deployment); otherwise the baked-in endpoint is used.
+    ///
+    /// `db_path` is a per-client location and `db_key` is a secret, so both are
+    /// caller-supplied — never baked into the library.
+    pub fn open(
+        transport: T,
+        db_path: impl Into<String>,
+        db_key: impl Into<String>,
+        registry_url: Option<&str>,
+    ) -> Result<(Self, Receiver<Event>), ClientError> {
+        let endpoint = registry_url.unwrap_or(REGISTRY_ENDPOINT);
+        ChatClientBuilder::new()
+            .ident(DelegateSigner::random())
+            .transport(transport)
+            .registration(HttpRegistry::new(endpoint))
+            .storage_config(StorageConfig::Encrypted {
+                path: db_path.into(),
+                key: db_key.into(),
+            })
+            .build()
+    }
+}
--- a/.github/scripts/mock-registry.py
+++ b/.github/scripts/mock-registry.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+"""Mock keypackage/account registry for the chat-cli CI smoketest.
+
+On startup the client registers its keypackage and account bundle. Publishing the
+bundle first fetches any existing record, so the stub answers:
+
+  * POST /v0/keypackage, POST /v0/account -> 200 (accept the write)
+  * GET  (any)                            -> 404 (nothing published yet)
+
+A 404 is what a fresh account looks like, which the client reads as "no existing
+record". This validates nothing — it only unblocks the smoketest; protocol-level
+behavior is covered by the workspace tests.
+"""
+
+from http.server import BaseHTTPRequestHandler, HTTPServer
+
+
+class Handler(BaseHTTPRequestHandler):
+    # Match the client's HTTP/1.1 requests so reqwest frames the response body.
+    protocol_version = "HTTP/1.1"
+
+    def _drain(self):
+        # Consume the request body so the client's request completes cleanly.
+        length = int(self.headers.get("Content-Length", 0))
+        if length:
+            self.rfile.read(length)
+
+    def _reply(self, status):
+        self.send_response(status)
+        self.send_header("Content-Length", "0")
+        self.end_headers()
+
+    def do_POST(self):
+        self._drain()
+        self._reply(200)
+
+    def do_GET(self):
+        self._drain()
+        self._reply(404)
+
+    def log_message(self, *args):
+        pass
+
+
+if __name__ == "__main__":
+    HTTPServer(("127.0.0.1", 18080), Handler).serve_forever()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,4 +74,15 @@ jobs:
       - name: Build chat-cli (logos-delivery)
         run: nix develop -c bash -c 'LOGOS_DELIVERY_LIB_DIR=./result/lib cargo build --release -p chat-cli'
       - name: Run chat-cli smoketest
-        run: nix develop -c ./target/release/chat-cli --name ci-test --smoketest
+        # The client registers against the keypackage/account registry on
+        # startup, so stand up a mock that accepts those calls. It runs on the
+        # system Python (no Nix needed); only chat-cli needs the dev shell.
+        run: |
+          python3 .github/scripts/mock-registry.py &
+          mock_pid=$!
+          trap 'kill "$mock_pid" 2>/dev/null || true' EXIT
+          for _ in $(seq 1 50); do
+            curl -s -o /dev/null http://127.0.0.1:18080/ && break
+            sleep 0.2
+          done
+          nix develop -c ./target/release/chat-cli --name ci-test --smoketest --registry-url http://127.0.0.1:18080
--- a/bin/chat-cli/src/main.rs
+++ b/bin/chat-cli/src/main.rs
@@ -9,8 +9,7 @@ use anyhow::{Context, Result};
 use clap::{Parser, ValueEnum};
 use crossbeam_channel::Receiver;
 use logos_chat::{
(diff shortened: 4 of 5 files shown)


crates/client/src/builder.rs
```diff
@@ -90,7 +106,11 @@ impl<I, T, R, S> ChatClientBuilder<I, T, R, S> {
 
 type Built<I, T, R, S> = Result<(ChatClient<I, T, R, S>, Receiver<Event>), ClientError>;
 
-// All four explicitly provided.
+/// `build()` exists only once every component has a concrete type. Any field
+/// still `Unset` (always at least the transport, which has no default) fails the
+/// bounds below, so an incomplete builder is a compile error rather than a
+/// runtime one. Start from [`ephemeral`](ChatClientBuilder::ephemeral) to fill the
+/// identity/registration/storage slots, then set the transport.
 impl<I, T, R, S> ChatClientBuilder<I, T, R, S>
```
jazzz, line 94: I like what you are trying to do here, by deleting the other build() impls however this goes against the benefits of using the "builder" pattern.

[Boulder] Implementors must not be required to provide every component on construction. For maintainability Developers only choose the components that they care about, and the library provides sensible defaults for the rest.

crates/client/src/builder.rs
```diff
@@ -22,18 +24,32 @@ pub struct ChatClientBuilder<I = Unset, T = Unset, R = Unset, S = Unset> {
...
-        Self::default()
+    /// A builder pre-filled with throwaway components — a random identity, an
+    /// ephemeral (in-memory) registry, and in-memory storage. Intended for tests,
+    /// examples, and local runs that need no persistence or network services;
+    /// override any slot with the matching setter. A complete entry point on its
+    /// own — only the transport is left to set before [`build`](ChatClientBuilder::build).
+    pub fn ephemeral() -> ChatClientBuilder<DelegateSigner, Unset, EphemeralRegistry, ChatStorage> {
+        Self::new()
+            .ident(DelegateSigner::random())
+            .registration(EphemeralRegistry::new())
```
jazzz: [Pebble] EphemeralRegistry requires clients be within the same process to connect. For developers this will be confusing why they cannot connect. Developers expect a EphemeralClient to be fully functioning rather than a test impl

crates/client/src/builder.rs
```diff
@@ -22,18 +24,32 @@ pub struct ChatClientBuilder<I = Unset, T = Unset, R = Unset, S = Unset> {
...
     }
-}
 
-impl ChatClientBuilder {
-    pub fn new() -> Self {
-        Self::default()
+    /// A builder pre-filled with throwaway components — a random identity, an
+    /// ephemeral (in-memory) registry, and in-memory storage. Intended for tests,
+    /// examples, and local runs that need no persistence or network services;
+    /// override any slot with the matching setter. A complete entry point on its
```
jazzz: [Sand] If intended for tests then why expose it at the primary developer entry point? Consider moving to integration tests, or tests module

crates/client/src/builder.rs
```diff
@@ -22,18 +24,32 @@ pub struct ChatClientBuilder<I = Unset, T = Unset, R = Unset, S = Unset> {
...
+    /// own — only the transport is left to set before [`build`](ChatClientBuilder::build).
+    pub fn ephemeral() -> ChatClientBuilder<DelegateSigner, Unset, EphemeralRegistry, ChatStorage> {
+        Self::new()
+            .ident(DelegateSigner::random())
```
jazzz: Using named configurations i think is a smart idea. 
However in this case it results in two undesirable paths for developers.

1) builder::new() + provide every single field, so it will build.
    - This breaks all clients if a new service is added to the client.
    - Requires developers to specify implementations they don't care about, which becomes failure points in the future
2) builder::<named_default>() + modifications.
    - Instantiates components which potentially have side effects, and are ultimately not used. 

The goal here is Developer choice + sensible defaults.

I would leave ChatClientBuilder as is and focus on providing named defaults via LogosChatClient

crates/client/src/logos.rs
```diff
@@ -0,0 +1,61 @@
...
+where
+    T: Transport + Send + 'static,
+{
+    /// Open a client on the Logos stack over `transport`, persisting to the
+    /// encrypted database at `db_path` unlocked with `db_key`. When `registry_url`
+    /// is `Some`, it overrides the preconfigured registry endpoint (e.g. a local
+    /// deployment); otherwise the baked-in endpoint is used.
+    ///
+    /// `db_path` is a per-client location and `db_key` is a secret, so both are
+    /// caller-supplied — never baked into the library.
```
jazzz, line 43: [Sand] this will be awkward for a while as all convos are missing persistence support, for the next month.

crates/client/src/logos.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub fn open(
+        transport: T,
+        db_path: impl Into<String>,
+        db_key: impl Into<String>,
+        registry_url: Option<&str>,
```
jazzz, line 48: [Sand] Consider removing registry_url.

For clients to interop developers need to be using the same registry service. Providing configuration here means that clients will be fragmented. 

In general we want to be very opinionated in our "opinionated" implementations - such that two developers using LogosChatClient can always communicate. Same Delivery, same registration, same account.

crates/client/src/logos.rs
```diff
@@ -0,0 +1,61 @@
...
+{
+    /// Open a client on the Logos stack over `transport`, persisting to the
+    /// encrypted database at `db_path` unlocked with `db_key`. When `registry_url`
+    /// is `Some`, it overrides the preconfigured registry endpoint (e.g. a local
+    /// deployment); otherwise the baked-in endpoint is used.
+    ///
+    /// `db_path` is a per-client location and `db_key` is a secret, so both are
+    /// caller-supplied — never baked into the library.
+    pub fn open(
+        transport: T,
```
jazzz, line 45: [Pebble] This should be opinionated about the transport being used. Developers should not have to choose this parameter, it should be configured for them if they choose the LogosChatClient

crates/client/examples/message-exchange/main.rs
```diff
@@ -6,13 +6,13 @@ fn main() {
     let bus = MessageBus::default();
     let reg = EphemeralRegistry::new();
 
-    let (mut saro, saro_events) = ChatClientBuilder::new()
+    let (mut saro, saro_events) = ChatClientBuilder::ephemeral()
         .transport(InProcessDelivery::new(bus.clone()))
         .registration(reg.clone())
         .build()
```
jazzz, line 12: [Dust] As a developer; Using Ephemeral to then potentially define `.storage(sql)` seems more confusing than building the client the way I want via `ChatClientBuilder::new()`

eg.

`ChatClientBuilder::new().transport(t).storage(sql)`
vs
`ChatClientBuilder::ephemeral().transport(t).storage(sql)`

Is it ephemeral or does it use, sql peristence?

crates/client/src/builder.rs
kaichaosun: The chat-store is not fully deployed, once deployed, we can update it to be be http registry.

crates/client/src/builder.rs
kaichaosun: Move to test modules sounds good, I kept it here in this PR to minimize the required changes. As I found it easier to progress with concentrate PRs.

crates/client/src/builder.rs
kaichaosun: > builder::new() + provide every single field, so it will build.

I think the `defaults()` and override is the way to go in most scenarios. And leaving `new()` as a fully controlled way to initiate the client with the "pain" to update along with library.

> Instantiates components which potentially have side effects, and are ultimately not used.

Good point, and I don't have a good solution yet, maybe we should defer the side effects. I also don't see it's a blocker as we are mostly using memory storage etc.

The refactor here is to solve **the bloated builder methods**, as it will be growing even worse when there are new dependency added. Macros is another way to try, but I'd like to try this solution first.

crates/client/src/logos.rs
kaichaosun: This is added here to facilitate locally running apps so before we have a production ready registry, it can be override for such apps. 
It will also be useful if we have logos-delivery broadcast capability for publish records in registry implement, so multiple instance can be supported.

crates/client/src/logos.rs
kaichaosun: Yeah, I suppose it should be fine as not many code depends on this path yet.

crates/client/src/logos.rs
kaichaosun: I agree, the current logos delivery is not in the proper place, and hard to instantiate it in `logos.rs`, needs another PR to first move logos delivery transport in extentions.

crates/client/src/builder.rs
kaichaosun: The new implementation actually fulfill this target as far as I can tell.
also checkout this comment https://github.com/logos-messaging/libchat/pull/151#discussion_r3478590879 for the reason why I remove the many builders.

crates/client/examples/message-exchange/main.rs
kaichaosun: The semantic looks a bit different,
`ChatClientBuilder::ephemeral().transport(t).storage(sql)` means you decide to go with default first and override later,
`ChatClientBuilder::new().transport(t).storage(sql).xyz()` are also supported, but devs needs to provide all the required ones.

Maybe we can rename `ephemeral` to `defaults` later when we have registry and persistence ready. I explicitly name is `ephemeral` to reflect its current usage and internal state.

crates/client/src/builder.rs
jazzz: > Good point, and I don't have a good solution yet, maybe we should defer the side effects.

That is the intention of the initial design. Initializing services will produce effects (registration, state initialization, logging, etc. ) As the services improve and become more production ready, the effects will increase. This interface is entirely focused on developers, so services like in memory storage will be replaced ASAP. 

 > The refactor here is to solve the bloated builder methods, as it will be growing even worse when there are new dependency added.
 
I don't see "bloated builder methods" as an issue that requires fixing at this moment. The code is relatively straight forward, and symmetric. Zero side-effect, compile-time type enforcement is worth it at this moment

> As I found it easier to progress with concentrate PRs.
Lets let it become an issue before attempting to resolve it

crates/client/src/builder.rs
jazzz: > The new implementation actually fulfill this target as far as I can tell.

I was imprecise with my language, as I was assuming a requirement of no side effects. 

Updated to: "Using a zero-side effect pathway like `::new()` Implementors must not be required to provide every component on construction. "

crates/client/src/builder.rs
jazzz: My concern is then, using http registry would not be ephemeral, as the KeyPackages are persisted beyond restarts. 

Ephemeral represents the current state well. However this is the primary entry point for developers. They do not want ephemeral services and messages. Public facing api structure need to be defined aspirationally, not based on the current feature sets.

crates/client/src/logos.rs
```diff
@@ -0,0 +1,61 @@
+//! The opinionated Logos client.
+//!
```
jazzz, line 2: Having LogosChatClient located within the "Generic client" seems like it will be full of friction as the objects have separate concerns. 

the "GenericClient will tend to be more generic/abstract" and the LogosChatClient will tend towards importing specific implementations and services. 

[Sand] LogosChatClient would benefit from being a separate crate, to keep imports separated. That way it can import LogosCore specific implementations and GenericClient could remain free of these implementation details

jazzz (changes requested):
My previous concerns still remain. 

- Blocker: Changing ChatClientBuilder to require an ephemeral base configuration has negative impacts on code maintenance and will be hard to undo in the future. 
   - This change does not seem necessary at this time, as it does not add any required functionality while removing beneficial properties.
   - The motivation of removing a few lines of code doesn't justify the increased maintenance costs in developer applications. 
- Forcing Ephemeral be the root of the builder pathway introduces fragmentation, which will become developer pain. 
   - The focus should be on creating an API that we want in the future( rather that what is available now), and working towards it with little to no breaking code changes in demos and developer applications.

crates/client/src/logos.rs
kaichaosun: Split LogosChatClient to its own crate looks too early, we can do it later. 
The landscape may also change as we move to production ready, and move ephemeral/in memory dependencies to test.

jazzz (approved):
Changes are looking much better

PR #153, opened 2026-06-25
chore: bump de-mls and adapt group_v2 to the new API

seemenkina:
Updates group_v2 to the latest de-mls.

de-mls reshaped its construction surface: key packages are now raw bytes plus an explicit member id (it no longer digs the id out of the package itself), the steward-list bounds moved into ConversationConfig, and create/join/add_member and the other driving methods regrouped their parameters. 

Behaviour is unchanged; this is keeping up with the dependency.

### DIFF
--- a/core/conversations/src/conversation/group_v2.rs
+++ b/core/conversations/src/conversation/group_v2.rs
@@ -7,16 +7,13 @@ use crate::{Content, WakeupService};
 use alloy::signers::local::PrivateKeySigner;
 use blake2::{Blake2b, Digest, digest::consts::U6};
 use chat_proto::logoschat::encryption::{EncryptedPayload, Plaintext, encrypted_payload};
-use de_mls::defaults::{
-    DefaultConsensusPlugin, DefaultPeerScoring, DefaultStewardList, InMemoryPeerScoreStorage,
-};
 use de_mls::protos::de_mls::messages::v1::{
     AppMessage as AppMessageProto, MemberWelcome, app_message,
 };
 use de_mls::{
-    ConsensusPlugin, ConsensusServiceFor, Conversation, ConversationConfig, ConversationEvent,
-    DeterministicStewardList, PeerScoringService, ScoringConfig, StewardListConfig,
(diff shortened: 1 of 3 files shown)


core/conversations/src/conversation/group_v2.rs
```diff
@@ -48,27 +45,16 @@ fn rand_app_id() -> Arc<[u8]> {
...
 /// random Ethereum consensus signer.
-fn make_consensus() -> ConsensusServiceFor<DefaultConsensusPlugin> {
-    ConsensusServiceFor::<DefaultConsensusPlugin>::new_with_components(
-        DefaultConsensusPlugin::new_storage(),
-        DefaultConsensusPlugin::new_event_bus(),
-        EthereumConsensusSigner::new(PrivateKeySigner::random()),
-        10,
-    )
+fn make_consensus() -> DefaultConsensusPlugin {
+    DefaultConsensusPlugin::new(EthereumConsensusSigner::new(PrivateKeySigner::random()))
```
jazzz, line 57: [?] In the future should the consensus signer match the MLS Signer? Or is it best to keep it separate?

jazzz, line 57: [?] It is safe to assume in the future that any implementor of ConsensusSignatureScheme would suffice here? Or is there a strict dependence on secp256K1 or the ethAddress format?

jazzz (approved):
Overall loving this direction. Smaller, cleaner, and easier to use.

core/conversations/src/conversation/group_v2.rs
seemenkina: Any signer that implement this trait: https://github.com/vacp2p/hashgraph-like-consensus/blob/48da54131328bb66a8f08dd3bf69c2727161f5d4/src/signing.rs#L46

seemenkina: No strict dependency anymore. At least I tried implementing general trait based on this idea, so if you find any contradictions let me know

PR #154, opened 2026-06-26
feat: re-export ChatStorage and make storage_config fallible

osmaczko:
Two small ergonomic changes to the `logos-chat` client crate, needed by callers
(logos-chat-module) that build an encrypted store and want to talk only to
`logos-chat` without depending on `libchat` directly.

- Re-export `ChatStorage` from `logos-chat` so callers can name the store type
  without a direct `libchat` dependency.
- `ChatClientBuilder::storage_config` now returns `Result<_, ChatError>` instead
  of `.expect()`, so an unopenable database surfaces as an error rather than
  aborting the process under `panic = "abort"`. Updates the chat-cli callers.

(Re-scoped from the earlier PrivateV1 message-delivery change, which is dropped:
logos-chat-module is moving to DirectV1, where senders are MLS-bound and the
PrivateV1 drop does not arise.)

### DIFF
--- a/bin/chat-cli/src/main.rs
+++ b/bin/chat-cli/src/main.rs
@@ -119,6 +119,7 @@ fn run<T: Transport>(transport: T, cli: &Cli) -> Result<()> {
             let (client, events) = ChatClientBuilder::new()
                 .transport(transport)
                 .storage_config(storage)
+                .map_err(|e| anyhow::anyhow!("{e:?}"))?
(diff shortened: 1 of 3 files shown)


jazzz:
FYI: PrivateV1 is End of life, and will be removed from the codebase in favor of DirectV1 once Client is compatible with InboxV2 registrations.

osmaczko:
Stale.

PR #155, opened 2026-06-27
fix: subscribe InboxV2 on routing_id, not the credential id

osmaczko:
InboxV2 gated inbound on the credential id (hex of the DelegateCredential TLV),
but a sender addresses the Welcome to the invitee's routing address, which the
account directory resolves to the device key HttpRegistry stores the key-package
under. The two differ, so the Welcome fell through to PayloadOutcome::Empty and
the invitee never joined. EphemeralRegistry hid this by keying key-packages on
the credential id.

This is routing-vs-credential, not account-vs-device: the TLV-wrapped credential
differs from the address even when account key == device key (testnet today), so
a client associating over HttpRegistry hits it; only the in-process
EphemeralRegistry path is unaffected.

Add IdentityProvider::routing_id() (defaults to id()); DelegateSigner derives it
from its credential (the account address once associated, else id());
Core::assemble subscribes InboxV2 on routing_id(). id() still backs the MLS
credential, member id, sender id, and decode_sender, so membership and
attribution are unchanged. Regression test direct_v1_welcome_routes_on_routing_id
over a device-key-keyed registry (as HttpRegistry) fails without the fix.

### DIFF
--- a/core/conversations/src/core.rs
+++ b/core/conversations/src/core.rs
@@ -112,7 +112,11 @@ where
         store: CS,
     ) -> Result<Self, ChatError> {
         let inbox = Inbox::new(&identity);
-        let ident_id = ident.id().clone();
+        // InboxV2 routes on routing_id, not the credential id: an
+        // account-associated delegate receives under its account address, which
+        // is what a peer resolves and addresses the Welcome to. The MLS identity
+        // below still wraps the raw `ident`, so the credential is unchanged.
+        let ident_id = ident.routing_id();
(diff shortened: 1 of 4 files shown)


osmaczko:
Superseded per the team decision that core should not be account-aware. This PR aligned the Welcome-routing mismatch by moving the inbox subscription *up* to the account address; #162 moves the sender side *down* to signer scope instead: accounts resolve to signer ids in the client layer, and one byte-string (the hex of the signer's verifying key) serves as registry key, inbox subscription, and Welcome routing target end to end.

PR #157, opened 2026-07-01
chore: gate embed p2p transport on cargo feature

kaichaosun:
Refactor on feature usage in embed p2p transport, use a consistent cargo feature, not env-dependent cfg

### DIFF
--- a/extensions/components/build.rs
+++ b/extensions/components/build.rs
@@ -9,17 +9,15 @@ fn main() {
         return;
     }
 
-    // The `embedded_p2p_delivery` feature compiles the FFI module, whose symbols
-    // resolve only against liblogosdelivery. The feature therefore *requires* the
-    // native library; if it can't be located, fail fast with actionable guidance
-    // rather than letting a downstream crate fail later with a confusing
-    // unresolved-import or link error.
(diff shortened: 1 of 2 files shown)


jazzz:
What is the plan for chat-cli? Does this change aim to remove the the transport choice within chat-cli and commit to using embedded λDelivery?


The existing approach, (supporting both cargo feature declaration as wells as envvars) was needed was to support the current setup of chat-cli, which allowed the delivery mechanism to be chosen at compile time via envvars.

kaichaosun:
Another PR after this change:
https://github.com/logos-messaging/libchat/compare/refactor-embed-transport-feature...more-transport-features?expand=1

I don't see any hard blocker to keep the env feature.

jazzz (approved):
> I don't see any hard blocker to keep the env feature.

If you mean that you want to remove the "env feature"  from the cli app then I don't see any issue with this change. 

I'd further suggest these changes for the future PR.
- removing the '--transport' flag all together. Embedded delivery needs to be compiled anyway. Remove the added complexity and simplify usage.
- `Logos-delivery` should be reserved for `logos-delivery` via `logos-core`. Use `embedded*` to refer to linking directly to the compiled library.

PR #159, opened 2026-07-02
feat: default transport for logos chat client

kaichaosun:
Makes logos-delivery the transport baked into LogosChatClient,
- LogosChatClient is no longer generic over transport, instead it takes `tcp_port` and `preset` and starts the EmbeddedP2pDeliveryService internally.
- components::EmbeddedP2pDeliveryService now implements Transport directly (in crates/client/src/logos.rs), so bin/chat-cli no longer needs its own P2pTransport newtype wrapper.
- update `chat-cli` to use logos chat client when fit

### DIFF
--- a/bin/chat-cli/Cargo.toml
+++ b/bin/chat-cli/Cargo.toml
@@ -9,9 +9,8 @@ path = "src/main.rs"
 
 [dependencies]
 # Workspace dependencies (sorted)
-components = { workspace = true , features = ["embedded_p2p_delivery"]}
 crossbeam-channel = { workspace = true }
-logos-chat = { workspace = true }
+logos-chat = { workspace = true, features = ["embedded-p2p-delivery"] }
 
 # External dependencies (sorted)
 anyhow = "1.0"
--- a/bin/chat-cli/src/main.rs
+++ b/bin/chat-cli/src/main.rs
@@ -9,29 +9,10 @@ use anyhow::{Context, Result};
 use clap::{Parser, ValueEnum};
 use crossbeam_channel::Receiver;
 use logos_chat::{
-    ChatClient, ChatStore, Event, IdentityProvider, LogosChatClient, RegistrationService, Transport,
+    ChatClient, ChatClientBuilder, ChatStore, DelegateSigner, Event, HttpRegistry,
+    IdentityProvider, LogosChatClient, NETWORK_PRESET, REGISTRY_ENDPOINT, RegistrationService,
+    StorageConfig, Transport,
 };
-
-use components::{EmbeddedP2pDeliveryService, P2pConfig};
-
-#[derive(Debug)]
-struct P2pTransport(EmbeddedP2pDeliveryService);
-
-impl logos_chat::DeliveryService for P2pTransport {
-    type Error = <EmbeddedP2pDeliveryService as logos_chat::DeliveryService>::Error;
-    fn publish(&mut self, envelope: logos_chat::AddressedEnvelope) -> Result<(), Self::Error> {
-        self.0.publish(envelope)
-    }
-    fn subscribe(&mut self, addr: &str) -> Result<(), Self::Error> {
-        self.0.subscribe(addr)
-    }
-}
-
-impl logos_chat::Transport for P2pTransport {
-    fn inbound(&mut self) -> crossbeam_channel::Receiver<Vec<u8>> {
-        self.0.inbound_queue()
-    }
-}
 
 use app::ChatApp;
 
@@ -62,9 +43,10 @@ struct Cli {
     db: Option<PathBuf>,
 
     // ── logos-delivery transport options ──────────────────────────────────────
-    /// logos-delivery network preset (e.g. `logos.dev`).
-    #[arg(long, default_value = "logos.dev")]
-    preset: String,
+    /// logos-delivery network preset (e.g. `logos.dev`). When omitted, the
+    /// preconfigured network preset is used.
+    #[arg(long)]
+    preset: Option<String>,
 
     /// TCP port for the embedded logos-delivery node.
     #[arg(long, default_value_t = 60000)]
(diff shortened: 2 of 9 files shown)


bin/chat-cli/src/main.rs
```diff
@@ -91,47 +73,66 @@ fn main() -> Result<()> {
...
+        // The file transport is a local-only path: it reuses the Logos service
+        // stack (delegate identity, HTTP registry, encrypted storage) but swaps
+        // the transport, so it builds a client directly instead of going through
+        // `LogosChatClient`.
         TransportKind::File => {
             let transport_dir = cli.data.join("transport");
             let transport = transport::file::FileTransport::new(&transport_dir)
                 .context("failed to create file transport")?;
-            run(transport, &cli)
-        }
```
jazzz, line 100: [Pebble] Remove `TransportKind::File` all together.

Keeping TransportKind::File will result in interop issues and broken conversations with other clients. The CliApp registers keybundles publically however the client is not routable, as it uses an isolated delivery service. When Account persistence is added this mean that any Account which uses File transport will be in a broken state.

bin/chat-cli/Cargo.toml
```diff
@@ -9,9 +9,8 @@ path = "src/main.rs"
 
 [dependencies]
 # Workspace dependencies (sorted)
-components = { workspace = true , features = ["embedded_p2p_delivery"]}
 crossbeam-channel = { workspace = true }
-logos-chat = { workspace = true }
+logos-chat = { workspace = true, features = ["logos-delivery"] }
```
jazzz: [Sand] We should reserve `logos-delivery` for the default use case which is logos-delivery+ logos-core. Use `embedded-p2p-delivery`

crates/client/src/lib.rs
```diff
@@ -1,17 +1,21 @@
 mod builder;
 mod client;
+mod config;
 mod delegate;
 mod delivery_in_process;
 mod errors;
 mod event;
+#[cfg(feature = "logos-delivery")]
 mod logos;
```
jazzz: [Sand] This feature name does not reflect it's purpose. Enabling `logos-delivery` compiles LogosChatClient, which is a bigger change than enabling logos core.

crates/client/src/lib.rs
```diff
@@ -1,17 +1,21 @@
...
 
 pub use builder::{ChatClientBuilder, Unset};
 pub use client::{ChatClient, Transport};
+pub use config::{NETWORK_PRESET, REGISTRY_ENDPOINT};
 pub use delegate::DelegateSigner;
 pub use delivery_in_process::{InProcessDelivery, MessageBus};
 pub use errors::ClientError;
 pub use event::{Event, MessageSender};
+#[cfg(feature = "logos-delivery")]
 pub use logos::LogosChatClient;
```
jazzz, line 19: [Pebble] Conditional compilation attribute is the wrong tool here.

This feature gate solely builds and exposes `LogosChatClient`. It does not alter code paths, or have any impacts on the existing `client` crate. In this case if conditional compilation is desired, use a separate crate - it's what its designed for. 

Conditional compilation attributes are useful when the core structure of an object needs to be altered. Or when controlling what is exposed through a support library to keep build sizes small.

Hiding the primary entry point for developers behind a feature gate, adds complexity and maintenance issues in applications `cargo.toml`. 99% of developers should exclusively be using LogosChatClient, and it should be the easiest path to use.

Exposing primary entrypoints via features:
- Means most developers will attempt to use Client api, as thats the easiest path.
- Makes documentation difficult.
- Its hard to link developers to code and apis

crates/client/src/logos.rs
```diff
@@ -1,52 +1,69 @@
...
+    /// is a per-client local resource, so all three are caller-supplied — never
+    /// baked into the library.
     pub fn open(
-        transport: T,
         db_path: impl Into<String>,
         db_key: impl Into<String>,
+        tcp_port: u16,
+        preset: Option<&str>,
         registry_url: Option<&str>,
     ) -> Result<(Self, Receiver<Event>), ClientError> {
```
jazzz, line 59: Removing transport from this constructor is smart. Love it. Being opinionated makes it easier on developers.

I'd like to see this be even more opinionated.

[Sand] Remove as many parameters as possible. Or use either a builder pattern or a mutable  DebugConfig object to reduce strain on developers. 

`registry_url`: Using anything other than the default would lead to fragmentation and inconsistent state within an openmls/accounts. Don'y allow developers to custom this value

`tcp_port`: Is there a need to make developers choose this? Can't the implementation choose a default?

`db_path`: I get why this is needed, but this is a footgun for developers. Favor user defined directories with deterministic filenames such that only one database ever exists for a signer app instance

`preset`: Developers choosing this value would corrupt state. KeyPackages are published to a shared location however commit messages are isolated to different networks.

At this point the key priority is maintaining interop between all clients and all applications. Developer customizations can come later when a need arises

jazzz: The high level issue is that removing  `transport` in the constructor results in developers having to make more choices, not less. 

Its easy enough to provide a function `create_default_transport()` which means that app developers can be hidden from these details. By flattening it, yet adding new transport specific parameters adds more cognitive load on developers rather than less.

jazzz (approved):
As LogosChatClient is intended to be the primary entry point for developers, I think there is some more work needed to make this developer ready. 

Specifically:
- Focusing on easier and simpler to use API, rather than more flexible options.
- Eliminating footguns that lead to Fragmentation, and interop failures.
- Moving to a separate crate to make imports, examples and documentation easier.

I won't expressly block this PR, as I understand this is a work in progress. However I don't think this PR brings us closer to our goals. I think its optimizing for options and flexibility rather than the app developer experience.

bin/chat-cli/src/main.rs
kaichaosun: Got it, but It's out of scope of this PR targeted.
The file transport for local test needs a better story, and it deserve its own focus.

crates/client/src/lib.rs
kaichaosun: Good point, though split to different crate is not good, as the current crate already called `logos-chat`, I need to think more before make any meaningful changes.

crates/client/src/logos.rs
kaichaosun: Future PRs will further target improve dev experience.

Issue #152, opened 2026-06-25
Group name and Description


jazzz:
Add configurable Groupname and Description to groupchats.

osmaczko:
Completed with https://github.com/logos-messaging/libchat/pull/183

PR #161, opened 2026-07-02
Add group description

jazzz:
## Problem

Groups lack a human readable identifier. With many groups the hex group_id blur into eachother making it difficult to know which group you are talking in. To avoid this, developers should be able to set a Name and description so that other participants  have the require context to communicate.  #152 

## Solution

This PR adds an Custom extension to the GroupContext. Which embeds a Payload into the encryption state. This approach ensures that every welcome contains the data immediately and is cryptographically bound to the group state.

## Notes
- This PR focuses solely on adding the required feature to the backend core, and uses a default value at the Core API. All callers and upstream APIs will be updated in a followup PR.
- This solution is not perfect, and will likely need to be revisited after some feedback has been gathered
- The custom Extension is not compatible with other MLS implementations. This is not a critical issue for this protocol, however worth noting for future discussion
- Waiting on an [upstream update](https://github.com/vacp2p/de-mls/pull/131), currently imports the feature branch to allow reviews. Will be removed before merging.

### DIFF
--- a/core/conversations/Cargo.toml
+++ b/core/conversations/Cargo.toml
@@ -18,7 +18,7 @@ storage = { workspace = true }
 alloy = "2.0"
 base64 = "0.22"
 chat-proto = { git = "https://github.com/logos-messaging/chat_proto", rev = "37ec98a151f6d50aab2905802ac0a896477e62ea" }
-de-mls = { git = "https://github.com/vacp2p/de-mls", branch = "main" }
+de-mls = { git = "https://github.com/vacp2p/de-mls", rev = "2c7a8669c1492c749c02efd2c5ac45e93e4926a3"} # Expose Mls Extensions (#131)
 double-ratchets = { path = "../double-ratchets" }
 hashgraph-like-consensus = "0.6.0"
 hex = "0.4.3"
--- a/core/conversations/src/conversation/mls_extensions.rs
+++ b/core/conversations/src/conversation/mls_extensions.rs
@@ -1,0 +1,120 @@
+use openmls::{
+    extensions::ExtensionType,
+    prelude::{
+        Capabilities,
+        tls_codec::{Deserialize, Error as TlsError, Serialize, Size, VLByteSlice, VLBytes},
+    },
+};
+use std::io::{Read, Write};
+
+use crate::types::ConvoMetadata;
+
+/// MLS extension type carrying our [`ConvoMetadata`]. In the private-use
+/// range (0xF000–0xFFFF) reserved by RFC 9420 for non-registered extensions.
+pub const GROUP_METADATA_EXTENSION_TYPE: u16 = 0xFF01;
+
+pub fn capabilities_with_group_metadata() -> Capabilities {
+    Capabilities::new(
+        None, // default protocol versions
+        None, // default ciphersuites
+        Some(&[ExtensionType::Unknown(GROUP_METADATA_EXTENSION_TYPE)]),
+        None, // default proposal types
+        None, // default credential types
+    )
+}
+
+/// Wire-format version of [`ConvoMetaInfo`], encoded as a `u16`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u16)]
+enum ConvoMetaInfoVersion {
+    V1 = 1,
+}
+
+impl TryFrom<u16> for ConvoMetaInfoVersion {
+    type Error = TlsError;
+
+    fn try_from(value: u16) -> Result<Self, Self::Error> {
+        match value {
+            1 => Ok(Self::V1),
+            other => Err(TlsError::DecodingError(format!(
+                "unknown ConvoMetaInfo version {other}"
+            ))),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct ConvoMetaInfo {
+    version: ConvoMetaInfoVersion,
+    name: String,
+    desc: String,
+}
+
+impl ConvoMetaInfo {
+    pub fn new(name: impl Into<String>, desc: impl Into<String>) -> Self {
+        Self {
+            version: ConvoMetaInfoVersion::V1,
+            name: name.into(),
+            desc: desc.into(),
+        }
+    }
+
+    pub fn to_extension_bytes(&self) -> Vec<u8> {
+        // TLS presentation-language encoding — matches the MLS stack's wire
+        // format. Writing to a Vec is infallible; the only error path is a
+        // field exceeding tls_codec's ~1 GiB length cap, unreachable here.
+        self.tls_serialize_detached()
+            .expect("ConvoMetaInfo serialization to Vec is infallible")
+    }
+
+    pub fn from_extension_bytes(bytes: &[u8]) -> Result<Self, TlsError> {
+        Self::tls_deserialize(&mut &bytes[..])
+    }
+}
+
+// Each field is encoded as a variable-length opaque (`opaque <V>`); `IdentId`
+// and `String` aren't `tls_codec` types, so we encode/decode their UTF-8 bytes.
+impl Size for ConvoMetaInfo {
+    fn tls_serialized_len(&self) -> usize {
+        (self.version as u16).tls_serialized_len()
+            + VLByteSlice(self.name.as_bytes()).tls_serialized_len()
+            + VLByteSlice(self.desc.as_bytes()).tls_serialized_len()
+    }
+}
+
+impl Serialize for ConvoMetaInfo {
+    fn tls_serialize<W: Write>(&self, writer: &mut W) -> Result<usize, TlsError> {
+        let mut written = (self.version as u16).tls_serialize(writer)?;
+        written += VLByteSlice(self.name.as_bytes()).tls_serialize(writer)?;
+        written += VLByteSlice(self.desc.as_bytes()).tls_serialize(writer)?;
+        Ok(written)
+    }
+}
+
+impl Deserialize for ConvoMetaInfo {
+    fn tls_deserialize<R: Read>(bytes: &mut R) -> Result<Self, TlsError> {
+        let version = ConvoMetaInfoVersion::try_from(u16::tls_deserialize(bytes)?)?;
+        let name = vl_string(bytes)?;
+        let desc = vl_string(bytes)?;
+        Ok(Self {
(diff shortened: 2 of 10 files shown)


core/conversations/Cargo.toml
```diff
@@ -18,7 +18,7 @@ storage = { workspace = true }
 alloy = "2.0"
 base64 = "0.22"
 chat-proto = { git = "https://github.com/logos-messaging/chat_proto", rev = "37ec98a151f6d50aab2905802ac0a896477e62ea" }
-de-mls = { git = "https://github.com/vacp2p/de-mls", branch = "main"}
+de-mls = { git = "https://github.com/vacp2p/de-mls", branch = "jazzz/expose_extensions"}
```
kaichaosun: [suggestion] We should pin the unpublished/untagged dependency to a commit, it makes the update explicit here. Pin it to a specific branch only updates in cargo.lock.

core/conversations/src/types.rs
```diff
@@ -66,3 +66,20 @@ impl AddressedEncryptedPayload {
         )
     }
 }
+
+#[derive(Debug)]
+pub struct ConvoMetadata {
+    pub owner: String,
```
kaichaosun: [question] 
1. why owner can not be typed? Looks pretty easy to misuse with a generic string.
2. what if it's a direct convo between two users, does it make sense to still have a owner, should each one has the same capability in this convo?

core/conversations/src/types.rs
```diff
@@ -66,3 +66,20 @@ impl AddressedEncryptedPayload {
...
+    pub name: String,
+    pub desc: String,
+}
+
+impl ConvoMetadata {
+    pub fn empty() -> Self {
```
kaichaosun: [suggestion] if `ConvoMetadata` can be omitted, it should be optional in usage, make the data empty looks not useful.

core/conversations/src/core.rs
```diff
@@ -224,7 +225,7 @@ impl<'a, S: ExternalServices + 'static> Core<S> {
         &mut self,
         participants: &[IdentIdRef],
     ) -> Result<ConversationId, ChatError> {
-        self.create_group_convo_v2(participants)
+        self.create_group_convo_v2(participants, "UnamedGroup", "")
```
kaichaosun: [request change] the hardcoded value doesn't make sense, if the api is not ready, the param should be optional.

core/conversations/src/conversation/mls_extensions.rs
```diff
@@ -0,0 +1,104 @@
...
+        Capabilities,
+        tls_codec::{Deserialize, Error as TlsError, Serialize, Size, VLByteSlice, VLBytes},
+    },
+};
+use shared_traits::{IdentId, IdentIdRef};
+use std::io::{Read, Write};
+
+use crate::types::ConvoMetadata;
+
+pub const GROUP_METADATA_EXTENSION_TYPE: u16 = 0xFF01;
```
kaichaosun, line 14: [request change] Please document the magic number and necessary background why such numbered extension type is needed.

core/conversations/src/conversation/mls_extensions.rs
```diff
@@ -0,0 +1,104 @@
...
+    version: u16,
+    name: String,
+    owner: IdentId,
+    desc: String,
+}
+
+impl ConvoMetaInfo {
+    pub fn new(name: impl Into<String>, owner: IdentIdRef, desc: impl Into<String>) -> Self {
+        Self {
+            version: 1,
```
kaichaosun: [suggestion] use enum for version is more extensible.

core/conversations/src/conversation/mls_extensions.rs
```diff
@@ -0,0 +1,104 @@
...
+            name: name.into(),
+            owner: owner.clone(),
+            desc: desc.into(),
+        }
+    }
+
+    pub fn to_extension_bytes(&self) -> Vec<u8> {
+        // TLS presentation-language encoding — matches the wire format used by
+        // the rest of the MLS stack, so no extra serializer is pulled in.
+        self.tls_serialize_detached().expect("serialization failed")
```
kaichaosun: [request change] avoid panic inside conversation.

core/conversations/src/conversation/mls_extensions.rs
```diff
@@ -0,0 +1,104 @@
...
+
+pub const GROUP_METADATA_EXTENSION_TYPE: u16 = 0xFF01;
+
+pub fn capabilities_with_group_metadata() -> Capabilities {
+    Capabilities::new(
+        None, // default protocol versions
+        None, // default ciphersuites
+        Some(&[ExtensionType::Unknown(GROUP_METADATA_EXTENSION_TYPE)]),
+        None, // default proposal types
+        None, // default credential types
```
kaichaosun, line 22: [suggestion] not sure why exactly it looks like this way, but the many `None`s make me feel that it's not in a proper abstraction.

core/conversations/src/conversation/mls_extensions.rs
```diff
@@ -0,0 +1,104 @@
...
+    )
+}
+
+#[derive(Debug, Clone)]
+pub struct ConvoMetaInfo {
+    version: u16,
+    name: String,
+    owner: IdentId,
```
kaichaosun: [suggestion] It's better to move the owner before name to keep it consistent in the codebase, currently some places name comes after owner, and here owner comes after name. 

The new method's param order applied too.

core/conversations/src/types.rs
jazzz: Owner removed

core/conversations/src/core.rs
jazzz: Updated PR description to be more clear.

```
This PR focuses solely on adding the required feature to the backend core, and uses a default value at the Core API. All callers and upstream APIs will be updated in a followup PR.
```
The proposed API is a `non optional` value passed in on construction. 

Without using a temporary default value this PR gets large and noisy quickly as it forces all the tests, clients, modules to be updated as well. 

The intention is to discuss refine, and merge the backend change. Once refined and merged, the callsites can be updated in a follow up.

Does this work?

core/conversations/src/conversation/mls_extensions.rs
jazzz: updated to :

```
/// MLS extension type carrying our [`ConvoMetadata`]. In the private-use
/// range (0xF000–0xFFFF) reserved by RFC 9420 for non-registered extensions.
pub const GROUP_METADATA_EXTENSION_TYPE: u16 = 0xFF01;
```

core/conversations/src/conversation/mls_extensions.rs
jazzz: Its an exclusively internal representation, but sure. Added in https://github.com/logos-messaging/libchat/pull/161/commits/0429ed76a5902a09d7820aa31b02947d3f0028e9

core/conversations/src/conversation/mls_extensions.rs
jazzz: Good eye. 

In this case its actually an *reasonably infallible function, but I should have Doc'd it better.

Ive updated the expect string as well as the comment to be clearer about the situation. What I don't want is to poison the entire call chain with Results for a function which cannot fail, and cannot be recovered from.

https://github.com/logos-messaging/libchat/pull/161/commits/9738a37fc6cb41658d30ff987707e1bc5157e955

Does that sound reasonable?

core/conversations/src/conversation/mls_extensions.rs
jazzz: This is a by product of OpenMLS using Options for its configuration. None is used to denote using the default value. 

My hope is that this Library can avoid such artifacts in our APIs

core/conversations/src/conversation/mls_extensions.rs
jazzz: Owner removed

core/conversations/Cargo.toml
jazzz: Thats a fair point. Since it only updates on cargo update .... why not make it explicit when it requires an explicit action. 

It makes the update slightly more verbose, but in general I agree

core/conversations/src/types.rs
jazzz: ConvoMetadata is not optional, however not supported by all ConvoTypes at this time.

Base on suggestion, I've:
- Internally the trait requires `Option<ConvoMetadata>`
   - Once all valid types support Metadata the option can be removed; simplifying code
- Externally Metadata returns a Result<ConvoMetadata, _> which errors when the convo_id is not found
   - In the interim it also fails when a type is not supported.
- ConvoMetadata::Empty has been removed. 

Does this work for you?

kaichaosun (approved):

core/conversations/src/conversation/mls_extensions.rs
```diff
@@ -0,0 +1,127 @@
...
+            ))),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct ConvoMetaInfo {
+    version: ConvoMetaInfoVersion,
+    name: String,
+    owner: IdentId,
```
osmaczko: [?] In theory, the owner should be all associated IdentIds (devices) of a given account. Given that, would it make more sense to have an account identifier here? Account is not really a conversation-level concern, though. Any ideas on how to solve this properly?

core/conversations/Cargo.toml
```diff
@@ -18,7 +18,7 @@ storage = { workspace = true }
 alloy = "2.0"
 base64 = "0.22"
 chat-proto = { git = "https://github.com/logos-messaging/chat_proto", rev = "37ec98a151f6d50aab2905802ac0a896477e62ea" }
-de-mls = { git = "https://github.com/vacp2p/de-mls", branch = "main"}
+de-mls = { git = "https://github.com/vacp2p/de-mls", rev = "2c7a8669c1492c749c02efd2c5ac45e93e4926a3"} # Expose Mls Extensions (#131)
```
osmaczko, line 21: Should we wait for it to be merged first?

osmaczko (approved):

core/conversations/Cargo.toml
jazzz: 💯

PR #162, opened 2026-07-02
fix: signer-scoped DirectV1 routing

osmaczko:
Core is no longer account-aware: the client resolves an account address
to signer ids via the account directory, and the signer's verifying-key
hex serves as registry key, inbox subscription, and Welcome routing
target end to end. The MLS credential stays the full id().

- GroupV2 reads the de-mls member id from the fetched key package and
  maps it to the signer id the welcome is delivered to.
- All account machinery (directory trait, bundle codec, resolution)
  moves out of core into logos-account; the RegistrationService
  supertrait and Core::account_directory() are gone, and the client
  holds its own directory handle.
- The account exposes functionality, never a signer: add_delegate_signer
  does the lamport upsert and signs internally.
- Every client acts for an account (ChatClientBuilder::new(account)).
  DelegateSigner is a pure keypair; the client composes the wire
  credential from the signer and the account, so the association is
  client state. addr() is the account address.
- resolve_device_ids fails fast (NotAnAccountKey / NoDeviceBundle /
  Directory) instead of falling back to treating an unresolved address
  as a signer id. LogosChatClient::open and chat-cli mint and publish a
  dev account each launch.
- EphemeralRegistry keys key packages by hex pubkey like HttpRegistry.

Supersedes #155 (routing_id).

### DIFF
--- a/core/account/src/account.rs
+++ b/core/account/src/account.rs
@@ -1,43 +1,183 @@
 use crypto::{Ed25519SigningKey, Ed25519VerifyingKey};
-use shared_traits::{IdentId, IdentIdRef};
 
-use libchat::IdentityProvider;
+use crate::directory::{AccountDirectory, SignedDeviceBundle, encode_bundle_payload};
 
-/// A Test Focused LogosAccount using a pre-defined identifier.
-/// The test account is not persisted, and uses a single user provided id.
+/// Failures updating an account's device bundle in the directory.
+#[derive(Debug, thiserror::Error)]
+pub enum AddDelegateSignerError {
+    #[error("directory: {0}")]
+    Directory(String),
+    #[error("directory returned a malformed device id")]
+    MalformedDeviceId,
+    #[error("directory returned a malformed device key")]
+    MalformedDeviceKey,
+}
+
+/// A Test Focused LogosAccount.
+/// The test account is not persisted.
 /// This account type should not be used in a production system.
 pub struct TestLogosAccount {
-    id: IdentId,
     signing_key: Ed25519SigningKey,
     verifying_key: Ed25519VerifyingKey,
 }
 
+impl Default for TestLogosAccount {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl TestLogosAccount {
-    pub fn new(explicit_id: impl Into<String>) -> Self {
+    pub fn new() -> Self {
         let signing_key = Ed25519SigningKey::generate();
         let verifying_key = signing_key.verifying_key();
         Self {
-            id: IdentId::new(explicit_id.into()),
             signing_key,
             verifying_key,
         }
     }
-}
 
-impl IdentityProvider for TestLogosAccount {
-    fn id(&self) -> IdentIdRef<'_> {
-        &self.id
-    }
-
-    fn display_name(&self) -> String {
-        self.id.to_string()
-    }
-
-    fn public_key(&self) -> &Ed25519VerifyingKey {
+    /// The account verifying key; its hex is the account address peers share.
+    pub fn public_key(&self) -> &Ed25519VerifyingKey {
         &self.verifying_key
     }
 
-    fn sign(&self, payload: &[u8]) -> crypto::Ed25519Signature {
-        self.signing_key.sign(payload)
+    /// The account address peers share: the hex of the verifying key.
+    pub fn address(&self) -> String {
+        hex::encode(self.verifying_key.as_ref())
+    }
+
+    /// Add `signer` (the delegate signer's verifying key) to this account's directory bundle.
+    ///
+    /// Fetches the current (verified) device set, adds the signer if absent,
+    /// bumps the lamport, re-signs, and publishes. Safe to call repeatedly:
+    /// an unchanged set is simply re-published, which also refreshes the
+    /// server's retention clock. The account signs internally; its key never
+    /// leaves this type.
+    pub fn add_delegate_signer<D: AccountDirectory>(
+        &self,
+        directory: &mut D,
+        signer: &Ed25519VerifyingKey,
+    ) -> Result<(), AddDelegateSignerError> {
+        // Start from the devices already registered so the account's other
+        // installations are preserved across the upsert.
+        let existing = directory
+            .fetch(&self.verifying_key)
+            .map_err(|e| AddDelegateSignerError::Directory(e.to_string()))?;
+        let (mut devices, next_lamport) = match existing {
+            Some(set) => {
+                let mut keys = Vec::with_capacity(set.devices.len() + 1);
+                for hex_id in &set.devices {
+                    let bytes: [u8; 32] = hex::decode(hex_id)
+                        .ok()
+                        .and_then(|b| b.try_into().ok())
+                        .ok_or(AddDelegateSignerError::MalformedDeviceId)?;
+                    let key = Ed25519VerifyingKey::from_bytes(&bytes)
+                        .map_err(|_| AddDelegateSignerError::MalformedDeviceKey)?;
+                    keys.push(key);
+                }
+                (keys, set.lamport + 1)
+            }
+            None => (Vec::new(), 0),
+        };
+
+        if !devices.iter().any(|d| d.as_ref() == signer.as_ref()) {
+            devices.push(signer.clone());
+        }
+
+        let payload = encode_bundle_payload(next_lamport, &devices);
+        let signature = self.signing_key.sign(&payload);
+        let bundle = SignedDeviceBundle {
+            account_pub: self.verifying_key.clone(),
+            payload,
+            signature,
+        };
+
+        directory
+            .publish(&bundle)
+            .map_err(|e| AddDelegateSignerError::Directory(e.to_string()))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::directory::{DeviceSet, verify_bundle};
+
+    use super::*;
+
+    /// Minimal in-test directory: stores the latest bundle, verifies on fetch.
+    #[derive(Debug, Default)]
+    struct FakeDir(Option<SignedDeviceBundle>);
+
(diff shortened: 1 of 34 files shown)


core/account/src/account.rs
```diff
@@ -41,3 +41,17 @@ impl IdentityProvider for TestLogosAccount {
         self.signing_key.sign(payload)
     }
 }
+
+// The test account holds its key locally, so it can act as its own signing
+// authority for device bundles.
+impl AccountAuthority for TestLogosAccount {
+    type Error = std::convert::Infallible;
```
jazzz: In general this is the correct place for account related functionality 👍

core/account/src/account.rs
```diff
@@ -41,3 +41,17 @@ impl IdentityProvider for TestLogosAccount {
...
+
+    fn account_pub(&self) -> &Ed25519VerifyingKey {
+        &self.verifying_key
+    }
+
+    fn sign(&self, payload: &[u8]) -> Result<crypto::Ed25519Signature, Self::Error> {
+        Ok(IdentityProvider::sign(self, payload))
+    }
```
jazzz: [!] Exposing access to an account signer is a bad design choice. Accounts should provide functionality (Add, Revoke, Rotate) and not provide dangerous surface areas.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -137,47 +136,38 @@ impl GroupV1Convo {
...
+        registry: &impl KeyPackageProvider,
+    ) -> Result<KeyPackage, ChatError> {
+        let retrieved = registry
+            .retrieve(signer.as_str())
+            .map_err(|e| ChatError::Generic(e.to_string()))?;
+        let Some(keypkg_bytes) = retrieved else {
+            return Err(ChatError::Protocol(format!(
+                "no keypackage for signer {signer}"
+            )));
+        };
```
jazzz, line 155: This looks good. 

Use the RegistryProvider to fetch KeyPackages based on the local identities.

core/conversations/src/conversation/group_v1.rs
```diff
@@ -137,47 +136,38 @@ impl GroupV1Convo {
...
         &mut self,
         content: &[u8],
         cx: &mut ServiceContext<S>,
     ) -> Result<(), ChatError> {
-        let sender_id = cx.mls_identity.id().as_str();
-        let reliable = cx.causal.on_send(&self.convo_id, sender_id, content);
+        // Causal-history sender hint: the signer routing id, the same
+        // identifier every other wire-level reference to this signer uses.
+        let sender_id = hex::encode(cx.mls_identity.public_key().as_ref());
+        let reliable = cx.causal.on_send(&self.convo_id, &sender_id, content);
```
jazzz: [Sand] These changes don't look necessary. Avoid altering casual if not needed

core/conversations/src/conversation/group_v2.rs
```diff
@@ -314,16 +328,17 @@ impl GroupV2Convo {
...
+        //    now). The welcome travels to the joiner's signer routing id, not
+        //    its de-mls member id.
         for evt in &events {
             if let ConversationEvent::WelcomeReady { welcome, .. } = evt {
                 for joiner in &welcome.joiner_identities {
-                    if let Some(i) = self.pending_invites.iter().position(|p| p == joiner) {
-                        self.pending_invites.remove(i);
-                        let name = String::from_utf8(joiner.clone()).map_err(ChatError::generic)?;
+                    if let Some(i) = self.pending_invites.iter().position(|(p, _)| p == joiner) {
+                        let (_, routing_id) = self.pending_invites.remove(i);
```
jazzz: [!] I'd have to check this, hard to tell if this is the correct Identifier to be using. 

This value should match the Id of the IdentityProvider which generated the keypackage

core/conversations/src/inbox_v2.rs
```diff
@@ -53,33 +48,33 @@ pub trait MlsProvider: OpenMlsProvider {
...
     })
     .map_err(ChatError::generic)
 }
 
-/// An PQ focused Conversation initializer.
-/// InboxV2 Incorporates an Account based identity system to support PQ based conversation protocols
-/// such as MLS.
+/// A PQ focused Conversation initializer.
+/// InboxV2 is signer-scoped: it receives invites under this installation's
+/// signer routing id, supporting PQ based conversation protocols such as MLS.
```
jazzz: [Sand] Weary of introducing yet another concept 'routing id'. IdentId could be improved, however feels like we should wait until it stablizes.

crates/client/src/directory.rs
```diff
@@ -0,0 +1,139 @@
...
+
+use crate::errors::ClientError;
+
+/// Add `device` to the account's directory bundle.
+///
+/// Fetches the current (verified) device set, adds the device if absent, bumps
+/// the lamport, re-signs with the account key, and publishes. Safe to call
+/// repeatedly — an unchanged set is simply re-published, which also refreshes
+/// the server's retention clock. Account key custody stays outside libchat:
+/// the injected [`AccountAuthority`] is only ever asked to sign.
```
jazzz: This is not the the correct place for this logic to exist

core/account/src/account.rs
```diff
@@ -1,43 +1,183 @@
...
+/// The test account is not persisted.
 /// This account type should not be used in a production system.
 pub struct TestLogosAccount {
-    id: IdentId,
     signing_key: Ed25519SigningKey,
     verifying_key: Ed25519VerifyingKey,
 }
 
+impl Default for TestLogosAccount {
+    fn default() -> Self {
```
kaichaosun, line 25: [nitpick] have a default value for account looks weird.

kaichaosun (approved):

core/account/src/account.rs
```diff
@@ -1,43 +1,183 @@
...
+    /// Add `signer` (the delegate signer's verifying key) to this account's directory bundle.
+    ///
+    /// Fetches the current (verified) device set, adds the signer if absent,
+    /// bumps the lamport, re-signs, and publishes. Safe to call repeatedly:
+    /// an unchanged set is simply re-published, which also refreshes the
+    /// server's retention clock. The account signs internally; its key never
+    /// leaves this type.
+    pub fn add_delegate_signer<D: AccountDirectory>(
+        &self,
+        directory: &mut D,
```
jazzz, line 59: [Sand] Making this function generic over D is problematic. The issue is that this implementation still relies on the keypackage provider to perform account operations.

core/conversations/src/service_context.rs
```diff
@@ -81,32 +79,11 @@ mod test_support {
...
-            _bundle: &SignedDeviceBundle,
-        ) -> Result<(), <Self as AccountDirectory>::Error> {
-            Ok(())
-        }
-
-        fn fetch(
-            &self,
-            _account: &Ed25519VerifyingKey,
-        ) -> Result<Option<DeviceSet>, <Self as AccountDirectory>::Error> {
+        fn retrieve(&self, _device_id: &str) -> Result<Option<Vec<u8>>, Self::Error> {
```
jazzz, line 86: 👍

crates/client/src/builder.rs
```diff
@@ -15,32 +16,34 @@ pub struct Unset;
...
+impl ChatClientBuilder {
+    /// Every client acts for an account, so the builder starts from its
+    /// address. It becomes the client's shareable address
+    /// ([`ChatClient::addr`]) and the account claim in the wire credential;
+    /// the account must endorse the signer in the directory for peers to
+    /// verify that claim.
+    pub fn new(account: impl Into<String>) -> Self {
         Self {
             ident: Unset,
+            account: account.into(),
```
jazzz, line 34: [Pebble] The account address is insufficient. The end result requires that the Identity type is replaced with an account object.

crates/client/src/builder.rs
```diff
@@ -184,12 +199,13 @@ where
 impl<T, R, S> ChatClientBuilder<Unset, T, R, S>
 where
     T: Transport + Send + 'static,
-    R: RegistrationService + Send + 'static,
+    R: RegistrationService + AccountDirectory + Clone + Send + 'static,
```
jazzz, line 202: [Sand] These are categorically two different services, and should eventually not be collocated.

extensions/components/src/contact_registry/ephemeral.rs
```diff
@@ -61,10 +59,12 @@ impl RegistrationService for EphemeralRegistry {
         identity: &dyn IdentityProvider,
         key_bundle: Vec<u8>,
     ) -> Result<(), <Self as RegistrationService>::Error> {
+        // Keyed by device id — the hex of the signer's verifying key — exactly
+        // like the HTTP registry, so tests exercise the deployed keying.
         self.key_packages
             .lock()
             .unwrap()
-            .insert(identity.id().to_string(), key_bundle);
+            .insert(hex::encode(identity.public_key().as_ref()), key_bundle);
```
jazzz, line 67: [Sand] Building non canoncial representations leads to bugs, and difficulty debugging. If a representation does not exist it should be added to the type.

The issue here is that IdentId is poorly designed and lacks clarity

jazzz (approved):
As a fix to continue developement I think its great, and have no blockers. 

This approach is generally in the right direction. 
- Removing AccountDirectory from core
- Cleaning up Identifiers.
- Introduction of a TestIdent

Things that need some more work in the future:
- Updates to the builder pattern will be hard to maintain
- Account crate needs to own all account operations
- Strongly typed Identifiers with canonical Identifiers will cut down on confusion

osmaczko:
@jazzz @kaichaosun thanks for the review. I’ll address your feedback in a follow-up. This is needed for `logos-chat-module`, hence the rush.

osmaczko:
Self-note: as agreed offline, @jazzz kindly offered to take over the suggested improvements (Thanks a lot for picking this up!).

PR #165, opened 2026-07-05
feat: group roster and honest DirectV1 conversation class

osmaczko:

## feat: expose a group's roster from libchat

  leaves.
- Unit tests cover the tolerant-vs-drop split; the three-member group
  integration test asserts the roster converges after create and after
  each add, and a solo group lists only its creator.

---

PR note: stacks on feat/groupv2-client (#164); the base commit c243c9c is that PR's head. libchat's ci.yml only runs on PRs to main, and #164's base is behind main, so this PR carries no CI until #164 merges (then rebase onto main). The change is verified locally (cargo test --locked -j1 + clippy --all-targets, both green) and transitively built by the module doctest (#44) and the chat-ui build (#31).


osmaczko:
Superseded by #164, which now carries the whole GroupV2 stack (client exposure, DirectV1 join-class fix, group roster with per-account dedup, registry retry) as one branch of atomic commits rebased on main.

PR #166, opened 2026-07-06
feat: separate embedded logos client

kaichaosun:
## Change

Split the client from the transport and draw the native-link boundary at a
crate, keeping `logos-chat` as the single entrypoint:

- **New crate `extensions/embedded-logos-delivery`** — the embedded
  logos-delivery transport service, moved out of `components` together with
  its FFI shims, `build.rs`, and `links` key. The service is renamed
  `EmbeddedP2pDeliveryService` → `EmbeddedLogosDelivery`, and the node
  defaults (network preset, TCP port) now live here with `P2pConfig`. The
  crate is excluded from the workspace's default members, so default
  builds/tests never need the native library. `components` no longer has any
  native dependency or feature.
- **`logos-chat` gains a transport-generic Logos stack**, compiled
  unconditionally: `LogosConfig` (db path/key + registry endpoint only — no
  transport settings), the `LogosChatClient<T>` alias, and
  `open_with_transport(config, transport)` for any injected `Transport`.
- **The embedded client is one concrete choice on top**, isolated in a
  `embedded` module behind the `embedded-logos-delivery` feature, which just
  switches on the optional dependency: the `Transport` impl for
  `EmbeddedLogosDelivery`, the `EmbeddedLogosClient` alias, and its
  `open(config, p2p_config)`. The feature also re-exports the transport types
  so consumers only ever depend on `logos-chat`.

Future transports (e.g. an FFI client) follow the same pattern: another named
alias over `LogosChatClient<T>`, plus an optional dependency only if it
carries a native cost.

## Consumers

- `chat-cli` enables the feature and opens the p2p path via
  `EmbeddedLogosClient::open`; its file-transport path reuses
  `open_with_transport` instead of hand-rolling the account/registry/storage
  wiring.
- CI's toolchain-only job now also excludes `embedded-logos-delivery`; the
  nix smoketest job still builds and runs the fully linked binary.


crates/client/src/logos.rs
```diff
@@ -85,21 +61,19 @@ impl LogosConfig {
...
-            preset: config.preset,
-            tcp_port: config.tcp_port,
-            ..Default::default()
-        })
-        .map_err(|e| ClientError::Transport(e.to_string()))?;
+/// account → device directory), and encrypted [`ChatStorage`] — generic over
+/// the transport `T`. Commit to a transport with
+/// [`open_with_transport`](Self::open_with_transport), or use a concrete
+/// alias like `EmbeddedLogosClient`.
+pub type LogosChatClient<T> = ChatClient<T, HttpRegistry, ChatStorage>;
```
jazzz: [Boulder] Remove generics from the simple interface. 

This change is counter productive to the intended goals. 

- Foolproof: LogosChatClient should ensure compatibility/interop between clients; Having a generic transport leads to network fragmentation and interaction failure.
 - Simplicity: LogosChatClient should be ruthlessly opinionated; Developers who want options should instantiate via the GenericClient. For everyone else it just works.

crates/client/src/logos.rs
```diff
@@ -85,21 +61,19 @@ impl LogosConfig {
...
 
+impl<T: Transport> LogosChatClient<T> {
+    /// Open a client on the Logos stack per `config` with the injected
+    /// transport, persisting to the encrypted database.
+    pub fn open_with_transport(
+        config: LogosConfig,
+        transport: T,
+    ) -> Result<(Self, Receiver<Event>), ClientError> {
```
jazzz: [Pebble] LogosChatClient is intended to be the "End" of the abstraction chain. This adds more layers of indirection, and more confusion

bin/chat-cli/Cargo.toml
```diff
@@ -11,7 +11,7 @@ path = "src/main.rs"
 # Workspace dependencies (sorted)
 crossbeam-channel = { workspace = true }
 logos-account = { workspace = true, features = ["dev"] }
-logos-chat = { workspace = true, features = ["embedded-p2p-delivery"] }
+logos-chat = { workspace = true, features = ["embedded-logos-delivery"] }
```
jazzz: [Sand] The primary import for developers looking for a "default" oepration should not require any features.

From a developer POV, this is extra complication and is frustrating. 
`cargo add logos-chat` should be all that is needed, and the example copied from the Repo readme should work.

crates/client/Cargo.toml
```diff
@@ -7,14 +7,15 @@ edition = "2024"
...
-embedded-p2p-delivery = ["components/embedded_p2p_delivery"]
+embedded-logos-delivery = ["dep:embedded-logos-delivery"]
 
 [dependencies]
 # Workspace dependencies (sorted)
 chat-sqlite = { workspace = true }
 components = { workspace = true }
 crossbeam-channel = { workspace = true }
 crypto = { workspace = true }
+embedded-logos-delivery = { workspace = true, optional = true }
```
jazzz: [Sand] As mentioned in #159 its my belief that LogosChatClient should be contained in its own crate so it is free to import what ever crates and features it wants. 

Client aka GenericClient ought to be completely agnostic to the trait impls.

I think its completely fair that package `logos-chat` ought to be where `LogosChatClient` lives. However I think that the existing implementation should be separate to maintain the separation that is already there. 

This is a `Sand` because its has no impact on the default developer path, and would be straight forward to migrate to split out. However I keep raising this point as its trivial to do now, and every new commit is becomes harder.

extensions/embedded-logos-delivery/src/lib.rs
```diff
@@ -115,21 +125,21 @@ impl WakuPayload {
     }
 }
 
-// ── EmbeddedP2pDeliveryService ──────────────────────────────────────────────────
+// ── EmbeddedLogosDelivery ──────────────────────────────────────────────────
```
jazzz, line 128: [Dust] This change seem unnescessary. I can see value in being clear about what P2P delivery is being used, however for most developers the specification will add more noise than clarity.

crates/client/src/lib.rs
```diff
@@ -1,22 +1,28 @@
 mod builder;
 mod client;
-mod config;
 mod delegate;
 mod delivery_in_process;
+#[cfg(feature = "embedded-logos-delivery")]
+mod embedded;
```
jazzz: [Dust] These compilation attributes are adding a lot of complexity.

extensions/embedded-logos-delivery/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "embedded-logos-delivery"
+version = "0.1.0"
+edition = "2024"
+links = "logosdelivery"
```
jazzz, line 5: [Pebble] I need to be convinced this is better than keeping components namespaced to he components crate. 

I'd agree that moving to its own crate has the benefit that the build infrastructure is easy to manage by default. However this doesn't appear to be a blocker.

The developer experience however appears significantly worse. When installing `logos-components` cargo lists available features (aka components) which are available for the advanced developer to pass to `ChatClientBuilder`. This allows developers to discover what is available and install their desired components to change behavior.

e.g. Current 
```
❯ cargo add components 
      Adding components to dependencies
             - embedded_p2p_delivery
```

In the future:
```
❯ cargo add logos-chat-components 
      Adding logos-chat-components to dependencies
             + logos_delivery
             + sqlite_message_store
             - embedded_p2p_delivery
             - ephemeral_message_store
             - keypackage_cache_client
```

It appears this change sacrifices developer experience for maintainer ease - What am I missing?

jazzz (commented):
In #163 I asked for strong justifications as to:
- why this approach makes LogosChat the easiest to use for new developers when compared to other options. 
- How we intend to minimize breaking changes in Apps during development.

Unfortunately this PR does not address my questions, and I still don't understand the reasoning behind these changes. 

As outlined in #159 my primary design goals are :
- Making the public entry point easier and simpler to use for developers, rather than more flexible options.
- Eliminating footguns that lead to Fragmentation, and interop failures.
- Making the entrypoint's imports, examples and documentation easier, by isolating to a separate crate. Reducing noise, and encouraging developers to become contributors.

I currently don't see how these PR's #159, #163, #166 contribute towards these goals.

extensions/embedded-logos-delivery/Cargo.toml
kaichaosun: I have experienced a few options,
1.  Separate crate for the whole Logos client, https://github.com/logos-messaging/libchat/compare/main...logos-separate-crate
2.  Narrow the feature gate, native code stays in components, https://github.com/logos-messaging/libchat/compare/main...logos-narrow-gate
3. this PR, split only the embedded logos-delivery into its own crate.

I proposed this solution instead of the others because:
- It isolates the embedded delivery properly. It has very different logic from the other components — a self-contained FFI wrapper (node thread, C callbacks, dylib install-name stamping) with no coupling to the pure-Rust registries and stores. Native-linking code getting its own crate is the **standard -sys-crate** convention.
- The feature-discovery DX you describe is preserved — it just lives on the entrypoint crate `logos-chat`. This is the same facade pattern as tokio's macros: features for discovery on the crate you install, crate boundaries for cost isolation underneath. Your future catalog is fully compatible — pure-Rust components can stay ordinary features; only native-linking ones become optional dependencies surfaced as features.
- The crate boundary is about scoping the native cost, not maintainer ease. Two things a feature inside components can't fix: the links = "logosdelivery" key and build.rs apply to the whole crate even with the feature off, so every components consumer (someone who only wants EphemeralRegistry) carries that machinery, and anything with --all-features (cargo test, docs tooling) hits the native link requirement. And cargo features are unified across the build graph — any crate anywhere enabling embedded_p2p_delivery imposes the link requirement on every build of components in that graph. A dependency edge scopes the cost to exactly the dependents who opted in.
- Why not the other two branches: option 1 splits the entrypoint, so developers need to know two crates — strictly worse DX. Option 2 fixes the client-side gating but leaves all consumers of components carrying the native build infrastructure, plus a two-hop feature chain (logos-chat/embedded-p2p-delivery → components/embedded_p2p_delivery).

So, discoverability is unchanged at the crate developers actually install, and the boundary moves the native-link cost from "everyone who touches components" to "exactly who wants the embedded node." Isn't this a better code architecture?

bin/chat-cli/Cargo.toml
kaichaosun: Agreed on the goal — cargo add logos-chat + readme example working is the right end state. 
But cargo can't build or fetch today (it comes from nix), the the feature isn't the complication; the undistributed native library is.

crates/client/src/logos.rs
kaichaosun: We have two use cases, 
1. logos-chat-module use logos chat client without embed logos delivery.
2. rust developers use logos chat client with embed or other integrated logos delivery.

The current `LogosChatClient` is for the 1st use case. And `EmbedLogosClient` is for the 2ed use case. Feel free to suggest a better name.

We may further consolidate the 1st and 2ed use case in the future, but seems not necessary for now. 
For 2ed user case, a new crate may be helpful, but it brings other issues like crate namings, clear scopes, which is risk too, but worth to explorer later.

crates/client/src/logos.rs
kaichaosun: Check this reply https://github.com/logos-messaging/libchat/pull/166#discussion_r3533992005

crates/client/src/logos.rs
jazzz: I think the two use-cases are accurate.

The 1st use case `logos-chat-module` uses GenericClient [directly](https://github.com/logos-co/logos-chat-module/blob/5c79789041c75bd940bdbc5720a2c9274df09b35/rust-lib/src/actions.rs#L130-L136). So I'm not concerned about that one. It could be optimized, but not a priority concern. 

> Feel free to suggest a better name.
`EmbedLogosClient` -> `LogosChatClient`. LogosChat client is the entry point for developers accessing this api, it has the recommended implementations.


>  but it brings other issues like crate namings, clear scopes,
My argument would be that now is the time to solve them. It will be harder to do later. 

`logos_chat` is the main entry point, feel free to rename the existing one to `generic_client` or something similar

kaichaosun: Done for split the crate, but still want to point out that, rush for the the splitting may make the underlining problems in the related code / product hard to surface early, for example 
- the unknown impact of embed logos delivery on developer experience
- not able to provide a clear uniform for similar use case.

The increased scope of related PR and embedded context in the PR's discussion has been surprise me continually, 
- great developer api needs to be taken care carefully and with a overall design, not embedded in a PR to resolve default configuration.
- before this PR, all the message I get is that logos-chat-module is our only customer, and in no short time we'll advocate a rust usage of libchat. Suddenly the focus shift from logos module to a new use case.

Overall, PRs should be constrained to its changes, ideas could be shared, but should not be extended to grow out of its scope.

jazzz: I agree here, on all points

extensions/embedded-logos-delivery/src/lib.rs
kaichaosun: As discussed, make it concrete until we are sure that p2p experience is good enough.

bin/chat-cli/Cargo.toml
jazzz: > But cargo can't build or fetch today (it comes from nix), the the feature isn't the complication; the undistributed native library is.

Are you sure that it doesn't build? I was under the impression it did, however I'd need to test that further

jazzz (approved):
I've been requested offline to allow this work to be merged as is. 

My reservations as described, still mostly stand.

However given priorities lie elsewhere currently and this change set does not make irrevocable commitments, I'm inclined to merge and close this work stream. Re-visiting this work should only occur once there has been a chance for alignment and planning about the best direction for the future.

bin/chat-cli/Cargo.toml
kaichaosun: I misspoke. Just verified with chat-cli build, It can be built with Cargo and Nix. Cargo calls Nix behind the scenes, which is convenient.

Issue #113, opened 2026-05-22
Cache Hydrated Conversations


jazzz:
## Problem

Currently the approach to managing conversation state is based on hydrating conversations from storage on every call. This has negative impacts on performance, readability, and ownership.

## Whats needed

- [ ] Clients or Inboxes should cache Conversations it has already loaded into memory. 
- [ ] Conversations state is persisted as needed.

jazzz:
Block on updates to the ServicesContext

PR #167, opened 2026-07-06
feat: GroupV2 through the threaded client, group roster, registry retry

osmaczko:
## feat: expose GroupV2 through the threaded client

GroupV2 (de-mls) conversations were reachable only from Core, and every
conversation ran the hardcoded millisecond timer profile in group_v2.rs
(20-150 ms freeze/consensus windows), which cannot survive real network
latency. Make them reachable through ChatClient (as DirectV1 already is), with
timing that holds up over a real network.

- ChatClient::create_group_conversation(accounts) and
  add_group_members(convo_id, accounts): resolve each account address to
  its endorsed signer ids through the client-held directory, the same
  resolution create_direct_conversation uses, and drive
  Core::create_group_convo / group_add_member.
- GroupV2 timing/policy is injectable: ServiceContext carries a
  de_mls::ConversationConfig (re-exported as GroupV2Config) defaulting
  to the de-mls library defaults; Core::set_group_v2_config and the
  builder's group_v2_config setter override it. The creator's phase
  durations reach joiners inside the welcome's ConversationSync, so the
  group runs the creator's phase timing.
- GroupV2Convo::add_member validates every member's key package before
  proposing any add, skips members de-mls would silently not propose
  (self, already in the group) instead of stranding a pending invite,
  and flushes opened proposals even on a mid-batch failure, so a failed
  batch cannot invite members behind the caller's back.
- The millisecond test profile moves into the test harnesses
  (integration_tests_core's TestHarness, crates/client/tests/group_v2.rs).
- New client-level tests: three accounts on the in-process transport
  create a group, a non-creator adds the third member, and messages fan
  out with directory-verified senders; a batch containing a member with
  no key package fails without inviting anyone.

## feat: expose a group's roster, deduped to one entry per account

  leaves.
- Because resolve_device_ids fans an account out to every endorsed
  device, an account whose devices all join surfaced once per device;
  group_members dedups by account, keeping the first-seen device as the
  account's representative. Members with no confirmed account stay
  individual, keyed by their unique device key.
- Unit tests cover the tolerant-vs-drop split and the per-account dedup;
  the three-member group integration test asserts the roster converges
  after create and after each add, and a solo group lists only its
  creator.

## feat: retry the registry on transient 5xx with backoff and jitter

The keypackage/account registry is reliable request-by-request but sheds concurrent bursts with a 5xx, so several instances registering at once each hard-failed on init. HttpRegistry's four calls now retry network errors and 5xx/429 with exponential backoff and full jitter (the jitter decorrelates concurrent publishers so their retries don't re-collide); 4xx and success return immediately. The total retry window is bounded to a few seconds.

## fix: mark InboxV2 key package last-resort so members can join multiple groups

A key package's init key is one-time-use: openmls deletes it after the first
welcome that consumes it. Each installation registers a single key package, so a
second group inviting the same member found no matching key package and rejected
the welcome with "welcome not addressed to this member", the flaky group add.

- Mark the InboxV2 key package as last-resort (and advertise the extension in
  the leaf capabilities, which key-package validation requires) so openmls
  retains the init key, letting one key package admit an installation to any
  number of groups.
- This reuses one init key for every join, trading per-join forward secrecy for
  membership that just works. A TODO at the publish site tracks the intended
  one-time key-package pool (the registry pops one per fetch, the client
  replenishes) with last-resort as the exhaustion fallback (#169).
- Regression tests: a member joining two groups (core harness) and two peers
  invited to several groups over the threaded client.

## fix: dedup list_conversations across the store and the in-memory cache

A DirectV1 join persists its conversation to the store and also caches it in
memory, so list_conversations saw it twice. It deduped with Vec::dedup, which
only drops consecutive repeats, over cached_convos' nondeterministic HashMap
order, so the duplicate survived whenever another cached conversation fell
between the two copies. A consumer counting conversations then saw a flaky count.

- Dedup through a set so a conversation held in both stores is listed once
  regardless of iteration order.
- Add a DirectV1-then-GroupV2 regression test, which also covers key-package
  reuse across conversation types.

### DIFF
--- a/core/conversations/src/conversation.rs
+++ b/core/conversations/src/conversation.rs
@@ -42,6 +42,10 @@ pub(crate) trait GroupConvo<S: ExternalServices>: Convo<S> + std::fmt::Debug + S
         cx: &mut ServiceContext<S>,
         members: &[IdentIdRef],
     ) -> Result<(), ChatError>;
+
+    /// Each current member's MLS leaf-credential content (hex-encoded), self
+    /// included.
+    fn members(&self) -> Result<Vec<Vec<u8>>, ChatError>;
 }
 
 pub(crate) trait Identified {
--- a/core/conversations/src/conversation/group_v2.rs
+++ b/core/conversations/src/conversation/group_v2.rs
@@ -316,6 +316,16 @@ where
         result.and(flushed)
     }
 
+    fn members(&self) -> Result<Vec<Vec<u8>>, ChatError> {
+        // Guarantee the local member is listed so callers see the full roster.
+        let mut members = self.conversation.members()?;
+        let self_id = self.conversation.member_id_bytes().to_vec();
+        if !members.contains(&self_id) {
+            members.push(self_id);
+        }
+        Ok(members)
+    }
+
     // fn conversation_state(&self) -> Result<ConversationState, ChatError> {
     //     Ok(self
     //         .conversation
--- a/core/conversations/src/conversation/group_v2.rs
+++ b/core/conversations/src/conversation/group_v2.rs
@@ -260,6 +260,17 @@ where
         service_ctx: &mut ServiceContext<S>,
         members: &[IdentIdRef],
     ) -> Result<(), ChatError> {
+        // Dedup the requested signers up front: an account can resolve to the
+        // same signer twice, or a caller can repeat one, and a duplicate would
+        // otherwise cost a redundant key-package fetch here and a second Add
+        // proposal for a member already being added in this batch.
+        let mut seen = std::collections::HashSet::new();
+        let members: Vec<IdentIdRef> = members
+            .iter()
+            .copied()
+            .filter(|m| seen.insert(m.as_str().to_string()))
+            .collect();
+
         // Fetch and validate every key package before proposing any add, so a
         // member with no key package fails the call before it opens proposals
         // for the others. Members are signer ids; the de-mls member id must
@@ -268,7 +279,7 @@ where
         // credential), so it is read from the fetched key package rather than
         // assumed equal to the signer id.
         let mut invites = Vec::with_capacity(members.len());
-        for member in members {
+        for member in &members {
             let kp_bytes = service_ctx
                 .registry
                 .retrieve(member.as_str())
@@ -285,16 +296,22 @@ where
             invites.push((member_id, member.to_string(), kp_bytes));
         }
 
-        // Record who WE invited before touching the conversation: after_op
-        // forwards a welcome only to joiners in pending_invites. Skip a member
-        // de-mls would silently not propose (self, or already in the group) —
-        // recording it would strand a pending invite that later matches a
-        // re-join and fires a duplicate welcome.
+        // pending_invites drives welcome delivery: after_op forwards a welcome
+        // only to a joiner recorded here. Record a member only if de-mls will
+        // actually propose its add — recording one it silently drops strands an
+        // entry that a later re-join can match, firing a spurious duplicate
+        // welcome. de-mls drops self and members already in the group; and since
+        // add_member only opens a proposal, the committed roster won't reflect a
+        // member added earlier in this same loop, so the set tracks those too.
+        // Seed it with the roster and self, insert as we go, and one check
+        // covers all three.
+        let mut roster: std::collections::HashSet<Vec<u8>> =
+            self.conversation.members()?.into_iter().collect();
+        roster.insert(self.conversation.member_id_bytes().to_vec());
+
         let mut result = Ok(());
         for (member_id, signer_id, kp_bytes) in invites {
-            if member_id == self.conversation.member_id_bytes()
-                || self.conversation.members()?.contains(&member_id)
-            {
+            if !roster.insert(member_id.clone()) {
                 continue;
             }
             self.pending_invites.push((member_id.clone(), signer_id));
--- a/core/conversations/src/conversation/group_v2.rs
+++ b/core/conversations/src/conversation/group_v2.rs
@@ -11,8 +11,7 @@ use de_mls::protos::de_mls::messages::v1::{
     AppMessage as AppMessageProto, MemberWelcome, app_message,
 };
 use de_mls::{
-    Conversation, ConversationConfig, ConversationEvent, PeerScoringService, ScoringConfig,
-    default_score_deltas,
+    Conversation, ConversationEvent, PeerScoringService, ScoringConfig, default_score_deltas,
     defaults::{DefaultConsensusPlugin, DefaultPeerScoring, InMemoryPeerScoreStorage},
 };
 use hashgraph_like_consensus::signing::EthereumConsensusSigner;
@@ -22,7 +21,6 @@ use openmls::prelude::{KeyPackageIn, OpenMlsProvider as _, ProtocolVersion};
 use prost::Message;
 use shared_traits::{IdentId, IdentIdRef};
 use std::sync::Arc;
-use std::time::Duration;
 use tracing::{info, instrument, warn};
 
 use crate::IdentityProvider;
@@ -56,21 +54,6 @@ fn make_scoring() -> DefaultPeerScoring {
 /// random Ethereum consensus signer.
 fn make_consensus() -> DefaultConsensusPlugin {
     DefaultConsensusPlugin::new(EthereumConsensusSigner::new(PrivateKeySigner::random()))
-}
-
-/// TEST-ONLY millisecond timers. de-mls deadlines are real wall-clock, so the
-/// default 60s timers never fire under fast virtual time. Production needs a
-/// real config injected from the caller, not these hardcoded values.
-fn demls_config() -> ConversationConfig {
-    ConversationConfig {
-        commit_inactivity_duration: Duration::from_millis(50),
-        freeze_duration: Duration::from_millis(20),
-        voting_delay: Duration::from_millis(30),
-        election_voting_delay: Duration::from_millis(30),
-        consensus_timeout: Duration::from_millis(150),
-        proposal_expiration: Duration::from_millis(2000),
-        ..ConversationConfig::default()
-    }
 }
 
 pub struct GroupV2Convo {
@@ -116,7 +99,7 @@ impl GroupV2Convo {
             &make_consensus(),
             make_scoring(),
             rand_app_id(),
-            demls_config(),
+            service_ctx.group_v2_config.clone(),
         )?;
         let convo = GroupV2Convo {
             convo_id,
@@ -147,7 +130,7 @@ impl GroupV2Convo {
             &make_consensus(),
             make_scoring(),
             rand_app_id(),
-            demls_config(),
+            service_ctx.group_v2_config.clone(),
         )?
         else {
             return Err(ChatError::generic("welcome not addressed to this member"));
@@ -277,13 +260,14 @@ where
         service_ctx: &mut ServiceContext<S>,
         members: &[IdentIdRef],
     ) -> Result<(), ChatError> {
-        // Record who WE invited before touching the conversation: after_op
-        // forwards a welcome only to joiners in pending_invites. Members are
-        // signer ids; the de-mls member id must match the id of the
-        // IdentityProvider that generated the key package (its MLS leaf
-        // credential content — de-mls matches members by credential), so it is
-        // read from the fetched key package rather than assumed equal to the
-        // signer id.
+        // Fetch and validate every key package before proposing any add, so a
+        // member with no key package fails the call before it opens proposals
+        // for the others. Members are signer ids; the de-mls member id must
+        // match the id of the IdentityProvider that generated the key package
+        // (its MLS leaf credential content — de-mls matches members by
+        // credential), so it is read from the fetched key package rather than
+        // assumed equal to the signer id.
+        let mut invites = Vec::with_capacity(members.len());
         for member in members {
             let kp_bytes = service_ctx
                 .registry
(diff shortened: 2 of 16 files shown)


core/conversations/src/conversation/group_v2.rs
```diff
@@ -298,17 +282,38 @@ where
...
-            self.conversation.add_member(
+            invites.push((member_id, member.to_string(), kp_bytes));
+        }
+
+        // Record who WE invited before touching the conversation: after_op
+        // forwards a welcome only to joiners in pending_invites. Skip a member
+        // de-mls would silently not propose (self, or already in the group) —
+        // recording it would strand a pending invite that later matches a
+        // re-join and fires a duplicate welcome.
+        let mut result = Ok(());
```
jazzz, line 312: [Sand] This doc string should be clearer about the problem. I had to read the code in order to understand, rather than the other way around.

Problem: Spurious welcomes can be sent if pending_invites contains members who cannot join the group.
Solution: Check that member is not already a member of the group

This is an interesting case that should be doc'd cleanly

core/conversations/src/conversation/group_v2.rs
```diff
@@ -298,17 +282,38 @@ where
...
+        for (member_id, signer_id, kp_bytes) in invites {
+            if member_id == self.conversation.member_id_bytes()
+                || self.conversation.members()?.contains(&member_id)
+            {
+                continue;
+            }
```
jazzz: [?] Should this check be occuring when `pending_invites` is processed in after_op? Seems like its still possible for this issue to occur. If so create an issue for later so bugs can be avoided.

core/conversations/src/conversation/group_v2.rs
```diff
@@ -298,17 +282,38 @@ where
...
                 &service_ctx.mls_provider,
                 &service_ctx.mls_identity,
                 &member_id,
                 &kp_bytes,
-            )?;
+            ) {
+                self.pending_invites.pop();
+                result = Err(e.into());
+                break;
+            }
```
jazzz, line 327: [!] Giving up on a failed `add_member` call seems reasonable. However curious if there are any error types that we'd want to attempt a retry in the future

core/conversations/src/conversation/group_v2.rs
```diff
@@ -298,17 +282,48 @@ where
...
         }
-        self.after_op(service_ctx)?;
-        Ok(())
+        // Flush even on a mid-loop failure: proposals already opened must be
+        // published and the wakeup re-armed, or they sit dormant until an
+        // unrelated frame drives the conversation.
+        let flushed = self.after_op(service_ctx).map(drop);
+        result.and(flushed)
```
jazzz, line 333: [?] I dont understand why `self.after_op(service_ctx)?;` doesn't work in this instance.

Doc comment appears to be about removing the early return from `self.conversation.add_member`

core/conversations/src/conversation.rs
```diff
@@ -42,6 +42,10 @@ pub(crate) trait GroupConvo<S: ExternalServices>: Convo<S> + std::fmt::Debug + S
         cx: &mut ServiceContext<S>,
         members: &[IdentIdRef],
     ) -> Result<(), ChatError>;
+
+    /// Each current member's MLS leaf-credential content (hex-encoded), self
+    /// included.
+    fn members(&self) -> Result<Vec<Vec<u8>>, ChatError>;
```
jazzz, line 48: [Sand] In the future this should be concretely typed so its easy to follow. Vec<Vec<u8>> doesn't help with type safety. I'll add this as part of the account work

core/conversations/src/core.rs
```diff
@@ -174,6 +175,12 @@ impl<'a, S: ExternalServices + 'static> Core<S> {
         self.pq_inbox.register(&mut self.services)
     }
 
+    /// Timing/policy for GroupV2 conversations created or joined after this
+    /// call. Existing conversations keep the config they were built with.
+    pub fn set_group_v2_config(&mut self, config: GroupV2Config) {
+        self.services.group_v2_config = config;
+    }
+
```
jazzz, line 183: [?] I hope this will go away once https://github.com/vacp2p/de-mls/pull/130 lands.

core/conversations/src/core.rs
```diff
@@ -293,6 +300,28 @@ impl<'a, S: ExternalServices + 'static> Core<S> {
...
+                .cached_convos
+                .get(convo_id)
+                .ok_or_else(|| ChatError::NoConvo(convo_id.to_string()))?;
+
+            match convo {
+                ConvoTypeOwned::Group(group_convo) => group_convo.members(),
+                ConvoTypeOwned::Direct(convo) => Err(ChatError::UnsupportedFunction(
+                    convo.id().into(),
+                    "List Members".into(),
+                )),
```
jazzz, line 317: [Sand] This will need to be implemented.

Becareful that Member does not mean account, it means delegateSigner. A DirectConvo will have multiple Members as each participant could have multiple installations

core/conversations/src/core.rs
```diff
@@ -293,6 +300,28 @@ impl<'a, S: ExternalServices + 'static> Core<S> {
...
+            }
+        } else {
+            let convo = self.load_group_convo(convo_id)?;
+            convo.members()
+        }
```
jazzz, line 322: [Future] This code path will be deprecated as PrivateV1 is no longer reachable.

core/conversations/src/core.rs
```diff
@@ -304,9 +333,13 @@ impl<'a, S: ExternalServices + 'static> Core<S> {
             convos.push(convo.to_string());
         }
 
-        // Conversations may use both storage mechanisms.
-        // Remove duplicates
-        convos.dedup();
+        // A conversation can live in both the store and the in-memory cache (a
+        // DirectV1 join persists to the store and is also cached), so drop
```
jazzz, line 337: [!] Any caching at this point is going to cause major UX issues, as only partial data is persisted. 
i.e the "Convo" is persisted but the backing DeMLS convo and account are not.

core/conversations/src/inbox_v2.rs
```diff
@@ -70,6 +71,10 @@ pub fn invite_user_v2<DS: DeliveryService>(
     .map_err(ChatError::generic)
 }
 
+/// A convo built from an InboxV2 invite, paired with the display class its
+/// invite type implies.
+type ClassifiedConvo<S> = (Box<dyn GroupConvo<S>>, ConversationClass);
```
jazzz, line 76: [Sand] Managing tuples is less than ideal. Consider adding a "class" concept to the Convo trait.

core/conversations/src/inbox_v2.rs
```diff
@@ -183,11 +195,22 @@ impl InboxV2 {
...
+        // no matching key package and reject the welcome ("welcome not addressed
+        // to this member"). Last-resort key packages are retained, so one admits
+        // the installation to any number of groups. Every key-package extension
+        // must be advertised in the leaf capabilities, hence LastResort there.
         let capabilities = Capabilities::builder()
             .ciphersuites(vec![CIPHER_SUITE])
-            .extensions(vec![ExtensionType::ApplicationId])
+            .extensions(vec![
+                ExtensionType::ApplicationId,
+                ExtensionType::LastResort,
```
jazzz, line 209: [Sand] Id like to avoid using this extension for as long as possible. Adding this should be a careful decision. Discussion in #169

core/conversations/src/service_context.rs
```diff
@@ -110,6 +115,7 @@ mod test_support {
                 causal: CausalHistoryStore::new(),
                 identity: Identity::new(name),
                 wakeup_service: NoopWakeups {},
+                group_v2_config: de_mls::ConversationConfig::default(),
```
jazzz, line 118: [Pebble] I don't think this can stay here long term. An issue is needed to resolve this issue cleanly.

crates/generic-chat/src/builder.rs
```diff
@@ -20,6 +20,7 @@ pub struct ChatClientBuilder<I = Unset, T = Unset, R = Unset, S = Unset> {
     transport: T,
     registration: R,
     storage: S,
+    group_v2: Option<GroupV2Config>,
```
jazzz, line 23: [~Boulder~ Sand] Exposing this at the client layer bleeds out abstraction layers, and will cause interop issues in the future.

Downgrading to a `Sand` because of the Builder pattern. Impact on developers/api is minimal. But still needs to be resolve soonly.

crates/generic-chat/src/client.rs
```diff
@@ -20,6 +21,19 @@ type ClientCore<T, R, S> = Core<(DelegateIdentity, T, R, ThreadedWakeupService,
 type AccountAddressRef<'a> = &'a str;
 type LocalSignerId = IdentId;
 
+/// A member of a group conversation's roster.
+///
+/// Shares [`MessageSender`]'s field semantics: `account` is set only when the
+/// member's credential claimed an account *and* the directory confirmed this
+/// device belongs to it. Unlike a message sender, an unconfirmable claim does
+/// not hide the member: it is cryptographically in the group, so it is listed
+/// by `local_identity` (its device) with `account: None`.
```
jazzz, line 30: [!] This is a great temporary solution. 

Account update will change this substantially, but this is clean fix for the current state

crates/client/src/client.rs
```diff
@@ -151,6 +169,54 @@ where
...
+
+        self.core
+            .lock()
+            .create_group_convo(&signer_refs)
+            .map_err(Into::into)
+    }
+
+    /// Add accounts' devices to an existing group conversation. Membership
+    /// changes are agreed and committed by the group asynchronously; the
+    /// joiners' welcomes go out once the add commits.
```
jazzz: > ... once the add commits.?

I'm not entirely sure what this means.

extensions/components/src/contact_registry/http.rs
```diff
@@ -266,6 +266,67 @@ fn decode_payload(payload: &[u8]) -> Option<(u64, &[u8])> {
     Some((timestamp_ms, &payload[8..]))
 }
 
+/// Retry budget for the registry's transient, load-induced 5xx/429 responses.
+/// The service is reliable request-by-request but sheds concurrent bursts, so a
+/// few backed-off retries let a request land once the burst clears. Sized to
+/// stay within chat_module's ~20s init IPC budget even at the worst-case sum.
```
jazzz: [!] If the infra is not working, lets create an issue to resolve it closer to the source.

jazzz (approved):
This PR provides many "Hole punch" solutions, but I don't see any strict blockers here.

No breaking impacts to public API, no issues with state desync as convo do not have persistence. 

My ask is to ensure that there are tickets opened for the larger issues(Pebbles+), but happy to merge when you are.

Notes::

I'm not yet onboard with exposing GroupV2Config generation via the service_ctx. This is being fixed in DeMLS. I'd eventually like to see GroupV2 provide a sensible default configuration for network operation - with tests being handled later. 

KeyPackage of last resort, is something that requires more consideration, and we'll probably revert later.

a destructive Accounts Update is incoming and will completely change that whole pathway.

For now I've tested it and it works nicely!

core/conversations/src/core.rs
```diff
@@ -140,6 +140,7 @@ where
                 causal,
                 identity,
                 wakeup_service,
+                group_v2_config: GroupV2Config::default(),
```
kaichaosun, line 143: [suggestion] this field is a config, but others is constructed service, is it possible to consume the `ground_v2_config` and produce a similar server and injected it here?

extensions/components/src/contact_registry/http.rs
kaichaosun: I agree, we should identify the issue asap, retry could tolerate some of the issues.

crates/generic-chat/tests/group_v2.rs
```diff
@@ -0,0 +1,332 @@
...
+use logos_chat::{
+    ChatClient, ChatClientBuilder, ConversationClass, DelegateSigner, Event, GroupV2Config,
+    InProcessDelivery, MessageBus,
+};
+
+/// Millisecond GroupV2 timers so the de-mls commit/consensus dance completes
+/// in test time; the library defaults wait 60s before committing an add.
+fn fast_group_v2_config() -> GroupV2Config {
+    GroupV2Config {
+        commit_inactivity_duration: Duration::from_millis(50),
```
kaichaosun, line 22: [request change] we should use const for such values in source code, so that some of the constants can be reused here.

crates/generic-chat/tests/group_v2.rs
```diff
@@ -0,0 +1,332 @@
...
+        } if got == content => Some(sender.account.as_ref().map(|a| a.as_str().to_string())),
+        _ => None,
+    })
+}
+
+/// A three-account group: saro creates it with raya, raya (a non-creator)
+/// adds pax, and a message from each member reaches both others with a
+/// directory-verified sender account.
+#[test]
+fn group_v2_three_members() {
```
kaichaosun, line 131:
```suggestion
fn three_members() {
```

core/conversations/src/conversation/group_v2.rs
```diff
@@ -298,17 +282,48 @@ where
...
+    }
+
+    fn members(&self) -> Result<Vec<Vec<u8>>, ChatError> {
```
kaichaosun, line 336: [suggestion] The function name should be clarified with what's the specific content to a member, for example `members_identity`.

kaichaosun (approved):

core/conversations/src/conversation/group_v2.rs
osmaczko: Reworded :heavy_check_mark:

core/conversations/src/conversation/group_v2.rs
osmaczko: https://github.com/logos-messaging/libchat/issues/170

core/conversations/src/conversation/group_v2.rs
osmaczko: All it buys is: on a double failure, surface the add error (the root cause) instead of the flush error.

core/conversations/src/core.rs
osmaczko: https://github.com/logos-messaging/libchat/issues/172

core/conversations/src/core.rs
osmaczko: https://github.com/logos-messaging/libchat/issues/171

core/conversations/src/core.rs
osmaczko: Already tracked by: https://github.com/logos-messaging/libchat/issues/135 and https://github.com/logos-messaging/libchat/issues/113

core/conversations/src/inbox_v2.rs
osmaczko: I do agree, though it is easy to change if one more parameter comes to play.

core/conversations/src/inbox_v2.rs
osmaczko: Let's keep for now, at least until #169 is resolved

core/conversations/src/service_context.rs
osmaczko: Tracked by: https://github.com/logos-messaging/libchat/issues/172

crates/generic-chat/src/builder.rs
osmaczko: Tracked by: https://github.com/logos-messaging/libchat/issues/172

crates/client/src/client.rs
osmaczko: Not the best wording, changed to:
```
    /// Add accounts' devices to an existing group conversation. The add is
    /// staged as an MLS proposal and merged by the group's next commit (driven
    /// asynchronously by the wakeup loop); each joiner's welcome is sent when
    /// that commit lands, not when this call returns.
```

extensions/components/src/contact_registry/http.rs
osmaczko: Tracked by https://github.com/logos-messaging/libchat/issues/173 - retry will be removed once this is completed

PR #168, opened 2026-07-07
feat: inject the GroupV2 time source and timing config

seemenkina:
GroupV2 (de-mls) conversations ran on real wall-clock time with hardcoded millisecond timers in group_v2.rs — timers that cannot survive real network latency, while production-scale values would have made every integration test sleep for real seconds (#164). de-mls now takes its clock from the caller (vacp2p/de-mls#130), so both the time source and the timing config become injectable.

- `GroupV2Clock` (`System` by default, `Mock(MockClock)` for tests) implements `de_mls::WallClock` and rides on `ServiceContext` next to the other services; `Core::set_group_v2_clock` overrides it. Every conversation deadline and consensus wire timestamp derives from this clock.
- The GroupV2 timing/policy config is injected the same way: `ServiceContext.demls_config` defaults to the de-mls library defaults and `Core::set_group_v2_config` overrides it. The hardcoded millisecond profile  is deleted from group_v2.rs and lives in the test harness, where it belongs.
- The test scheduler's `advance_time` no longer sleeps: all clients share one `MockClock` advanced in lockstep with the scheduler's virtual `now`, so protocol deadlines and wakeups finally run on the same timeline.
- de-mls is pinned to its PR branch and `hashgraph-like-consensus` bumped to 0.6 (its API takes `now` per call).

Before merge we should bump de-mls link to the main


core/conversations/src/conversation/group_v2.rs
```diff
@@ -32,6 +32,29 @@ use crate::{
...
+    System,
+    Mock(MockClock),
+}
+
+impl WallClock for GroupV2Clock {
+    fn now(&self) -> Duration {
+        match self {
+            GroupV2Clock::System => SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap_or_default(),
```
jazzz, line 52: [!] SystemTime allows time to go backwards will that be an issue?

> this time measurement is not monotonic. This means that you can save a file to the file system, then save another file to the file system, and the second file has a SystemTime measurement earlier than the first. In other words, an operation that happens after another operation in real time may have an earlier SystemTime!

ref: https://doc.rust-lang.org/std/time/struct.SystemTime.html

core/integration_tests_core/src/test_client.rs
```diff
@@ -291,6 +294,20 @@ impl TestHarness<4> {
...
+        commit_inactivity_duration: Duration::from_millis(50),
+        freeze_duration: Duration::from_millis(20),
+        voting_delay: Duration::from_millis(30),
+        election_voting_delay: Duration::from_millis(30),
+        consensus_timeout: Duration::from_millis(150),
+        proposal_expiration: Duration::from_millis(2000),
+        ..GroupV2Config::default()
+    }
+}
+
```
jazzz, line 310: [?] For my own knowledge; if the clock service can control time, can a future TestClock be used to advance time faster so the default config can work for tests as well?

jazzz (approved):
Looks good for an initial integration. 

cc: @osmaczko this will eventually need to be moved to the components crate, but its a low priority at this point

core/integration_tests_core/src/test_client.rs
osmaczko: Interesting idea

core/conversations/src/conversation/group_v2.rs
```diff
@@ -32,6 +32,29 @@ use crate::{
...
+/// The de-mls time source: every conversation deadline (freeze windows,
+/// consensus timeouts, auto-votes) and consensus wire timestamp is measured
+/// against this clock. Production runs on system time; tests share one
+/// `MockClock` with the harness scheduler so virtual time moves the
+/// protocol's timers.
+#[derive(Debug, Clone, Default)]
+pub enum GroupV2Clock {
+    #[default]
+    System,
+    Mock(MockClock),
```
osmaczko, line 44: [Sand] I think it would be better to keep production code as unaware of tests as possible. Mocks should ideally only be accessible from tests.

osmaczko (approved):
Thanks 🙏 

> Before merge we should bump de-mls link to the main

Yes, pease.

jazzz:
Merging this PR is blocking development. If there are any followup tasks lets get them completed, in a follow up

PR #177, opened 2026-07-12
feat: surface a roster-changed event when group membership changes

osmaczko:
A group's roster changed silently. An add merges on the steward's commit-inactivity timer, and the members that receive its commit apply it, but neither surfaced any observation: Core::wakeup returned () and the client worker discarded it, and GroupV2Convo mapped only chat messages into a ConvoOutcome, so a commit produced an empty outcome. An app could learn a group grew only by re-selecting the conversation, a manual refresh, or a later message from the new member.

de-mls already reports the change: CommitApplied (adds and removes) and WelcomeReady (adds) both fire, on every member, when a commit merges. Drain them into the observation and emit a new event.

- ConvoOutcome gains members_changed, set by GroupV2Convo when a poll cycle's drained de-mls events include CommitApplied or WelcomeReady. It rides alongside content, the same way a protocol-only frame already yields content: None.
- Convo::wakeup returns a ConvoOutcome instead of (), mirroring handle_frame, so the steward's own timer-driven commit is observable. Kinds with no timers return ConvoOutcome::empty; Core::wakeup and the worker translate it through the same events_from_inbound path inbound payloads use.
- New Event::ConversationMembersChanged { convo_id }: the app re-fetches group_members. Fires on every member a commit reaches, so the inviter sees its own add land and existing members see later joins.

---
Part of the member-add UX stack. libchat (this PR) adds the event; consumed by logos-co/logos-chat-module and logos-co/logos-chat-ui.

### DIFF
--- a/core/conversations/src/outcomes.rs
+++ b/core/conversations/src/outcomes.rs
@@ -1,7 +1,7 @@
 //! Observations a single inbound payload produces.
 //!
 //! - [`ConvoOutcome`] — an optional [`Content`] on a single existing
-//!   conversation.
+//!   conversation, plus whether a commit changed its membership.
 //! - [`InboxOutcome`] — a newly observed conversation, optionally with an
 //!   initial [`ConvoOutcome`].
 //! - [`PayloadOutcome`] — the union of the above, plus `Empty`.
@@ -22,6 +22,7 @@ pub struct Content {
 pub struct ConvoOutcome {
     pub convo_id: ConversationId,
     pub content: Option<Content>,
+    pub members_changed: bool,
 }
 
 impl ConvoOutcome {
@@ -29,6 +30,7 @@ impl ConvoOutcome {
         Self {
             convo_id,
             content: None,
+            members_changed: false,
         }
     }
 }
--- a/crates/generic-chat/src/event.rs
+++ b/crates/generic-chat/src/event.rs
@@ -38,6 +38,10 @@ pub enum Event {
         content: Vec<u8>,
         sender: MessageSender,
     },
+    /// A commit changed a conversation's membership.
+    ConversationMembersChanged {
+        convo_id: Arc<str>,
+    },
     InboundError {
         message: String,
     },
--- a/core/conversations/src/conversation.rs
+++ b/core/conversations/src/conversation.rs
@@ -34,7 +34,9 @@ pub(crate) trait Convo<S: ExternalServices>: Identified + Send {
         enc: EncryptedPayload,
     ) -> Result<ConvoOutcome, ChatError>;
 
-    fn wakeup(&mut self, service_ctx: &mut ServiceContext<S>) -> Result<(), ChatError>;
+    /// Advances any time-driven protocol work (de-mls consensus deadlines) and
+    /// reports what it observed, mirroring [`Self::handle_frame`].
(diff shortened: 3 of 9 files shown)


core/conversations/src/outcomes.rs
```diff
@@ -22,13 +22,15 @@ pub struct Content {
 pub struct ConvoOutcome {
     pub convo_id: ConversationId,
     pub content: Option<Content>,
+    pub members_changed: bool,
```
osmaczko, line 25: "_Premature optimization is the root of all evil_". I considered a typed `added`/`removed` delta, but a full re-fetch is easier to maintain: it stays self-correcting against e.g. dropped events, whereas a delta would add extra logic like credential-to-address resolution and dedup.

jazzz: > "Premature optimization is the root of all evil"

[!] In general I like the simplicity of the approach. However I'm not sure the issues you describe are possible.

> it stays self-correcting against e.g. dropped events, 

Membership change always results in a new epoch. If a group state changed is missed the group will be broken. It's not possible for a member to drop a membership change and still be a member of the group

> credential-to-address resolution and dedup.

With the current architecture this needs to happens at the client - not the core.

crates/generic-chat/src/event.rs
```diff
@@ -38,6 +38,10 @@ pub enum Event {
         content: Vec<u8>,
         sender: MessageSender,
     },
+    /// A commit changed a conversation's membership.
+    ConversationMembersChanged {
+        convo_id: Arc<str>,
+    },
```
jazzz, line 44: [!] I think the approach of keeping this simple for now is a smart move. This will need future work to optimize for developers, however lets get some datapoints first

jazzz (approved):
Adding an Outcome to Wakeup handler is a good change.

I like how this is forward progress without trying to "solve the whole world"

core/conversations/src/outcomes.rs
kaichaosun: [suggestion] It seems `members_changed` is not extensible enough with a boolean value.
Wondering if it's possible to return the root of mls ratchet tree into something like `ConvoState`, then notify with a ConvoStateUpdate event when new root arrived.

kaichaosun (approved):

core/conversations/src/outcomes.rs
osmaczko: > > it stays self-correcting against e.g. dropped events,
> 
> Membership change always results in a new epoch. If a group state changed is missed the group will be broken. It's not possible for a member to drop a membership change and still be a member of the group

Sorry, I should have been more precise. I was referring to the client → application flow. If the application drops an added/removed delta event for any reason, it may end up with an incorrect member list. What I meant is that explicitly refetching the member list on a members_changed event would be less error-prone.

> With the current architecture this needs to happens at the client - not the core.

Yeah, I tagged the wrong line of code. I should have tagged the event on the client. It is the same issue as above, i.e. my comment was made at the wrong layer.

osmaczko: > [suggestion] It seems `members_changed` is not extensible enough with a boolean value. Wondering if it's possible to return the root of mls ratchet tree into something like `ConvoState`, then notify with a ConvoStateUpdate event when new root arrived.

Let's KISS for now. It is trivial to update when there is a use-case.

PR #178, opened 2026-07-13
Isolate delivery library

jazzz:
## Problem

`embedded-logos-delivery` is not usable as a crate, as it assumed it was going to be used as a Libchat::DeliveryService which is not always true. LogosDelivery has no plans to support a rust binding, which makes using delivery for other tools.

## Solution
Split the embedded logos-delivery transport into two layers:

- logos-delivery (new crate at `extensions/logos-delivery-rust/` — owns everything tied to the native node: the raw FFI (sys), the safe synchronous FFI wrapper (wrapper::LogosNodeCtx), and a generic, transport-agnostic ThreadedDeliveryWrapper<T> that runs the node on a dedicated thread. It re-exports P2pConfig, DeliveryError, WakuEvent, and ReceivedMessage.
- embedded-logos-delivery — now a thin adapter over that wrapper. It keeps only the delivery-specific concerns: mapping AddressedEnvelope.delivery_address ↔ the /logos-chat/1/{addr}/proto content topic, filtering to the /logos-chat/1/… prefix, and decoding payloads to raw bytes.
- The links = "logosdelivery" native-link declaration moves to the new crate accordingly.

### Key changes
- Generic wrapper. ThreadedDeliveryWrapper<T> takes a caller-supplied FnMut(WakuEvent) -> Option<T> mapper that runs on the callback thread, so filtering and decoding happen inline and the inbound queue yields exactly the type the caller wants. embedded-logos-delivery instantiates it as ThreadedDeliveryWrapper<Vec<u8>>.
- Real subscribe/unsubscribe. Node operations are now modeled as a NodeOp enum (Publish / Subscribe / Unsubscribe) serialized through the node thread. DeliveryService::subscribe routes to the node instead of being a no-op, and a matching unsubscribe is added.
- Workspace wiring. Adds extensions/logos-delivery-rust as a workspace member and the logos-delivery path dependency.


extensions/embedded-logos-delivery/src/lib.rs
```diff
@@ -1,325 +1,91 @@
...
-    payload: WakuPayload,
-}
-
-/// Untagged union that handles both payload representations.
-#[derive(Debug, serde::Deserialize)]
-#[serde(untagged)]
-enum WakuPayload {
-    Base64(String),
-    Bytes(Vec<u8>),
+pub fn content_topic_for(delivery_address: &str) -> String {
```
kaichaosun: [nitpick] define `CHAT_TOPIC_PREFIX` first and reuse it here.

extensions/embedded-logos-delivery/src/lib.rs
```diff
@@ -1,325 +1,91 @@
 //! The embedded logos-delivery transport service.
```
kaichaosun, line 1: [suggestion] The thin wrapper in this crate can be moved to logo-chat, make it a separate crate looks like a overkill.

extensions/logos-delivery-rust/src/threaded.rs
```diff
@@ -0,0 +1,360 @@
...
+    {
+        // discv5UdpPort defaults to 9000 in libwaku, so a second instance with
+        // a distinct --port still collides on UDP. Bind it to tcp_port so a
+        // single --port knob keeps both ports distinct across instances.
+        let config_json = serde_json::json!({
+            "logLevel": cfg.log_level,
+            "mode": "Core",
+            "preset": cfg.preset,
+            "tcpPort": cfg.tcp_port,
+            "discv5UdpPort": cfg.tcp_port,
```
kaichaosun: [request change] this looks confusing to assign tcp port to a udp field. Better to give the port specific meaning like what libwaku, adding an extra config item here looks fine as we are harden the logos-delivery integration.

extensions/embedded-logos-delivery/src/lib.rs
jazzz: I'm not sure I am following here. As I see the history here.

1) The trait implementations(DeliveryService, Transport) were moved from a high level app to components so they could be re-used. https://github.com/logos-messaging/libchat/pull/148
2) In #166 you wanted to separate it to its own crate ref: https://github.com/logos-messaging/libchat/pull/166#discussion_r3533497699 

Now I'm understanding the suggestion is to move the trait implementations to logos-chat which would undo this, by making a transport impl un-accessible again.

I'm worried this is going in circles.

extensions/embedded-logos-delivery/src/lib.rs
jazzz: sure. 

[!] All topic definitions need to be defined in core to maintain interop, so this won't last long but lets keep it clean.

extensions/embedded-logos-delivery/src/lib.rs
kaichaosun: > Now I'm understanding the suggestion is to move the trait implementations to logos-chat

My thoughts on these changes are that we have separated logos-delivery, and clear feature gates in definition and its usage. The trait can be moved to the places that meets this target, make it an extra crate looks not the best in current shape, since:
- the logos-delivery-rust is splited
- logos-chat has been splited

This is a general suggestion, I'm happy to take any path you decide on it.

kaichaosun (approved):

extensions/logos-delivery-rust/src/threaded.rs
jazzz: removed all mentions of tcp to avoid confusion

extensions/embedded-logos-delivery/src/lib.rs
jazzz: Very much agree, that embedded-logos-delivery should not be a crate.

PR #183, opened 2026-07-16
feat: expose group name and description through the threaded client

osmaczko:
A GroupV2 conversation carries a shared name and description in its MLS group context, set at creation and delivered to every joiner in the welcome. The threaded client could create groups but neither set that metadata nor read it back, leaving both fields unreachable through the client API.

create_group_conversation now takes name and desc and creates the group with them via create_group_convo_v2; either may be empty for an unnamed group. A new group_metadata getter reads a group's metadata back, erroring for a direct conversation or a legacy group that carries none. ConvoMetadata is re-exported from the crate root and from logos-generic-chat so callers can name the returned type.

### DIFF
--- a/crates/generic-chat/src/client.rs
+++ b/crates/generic-chat/src/client.rs
@@ -6,8 +6,8 @@ use components::{ThreadedWakeupService, WakeupEvent};
 use crossbeam_channel::{Receiver, Sender, select};
 use crypto::Ed25519VerifyingKey;
 use libchat::{
-    ConversationId, ConvoOutcome, Core, DeliveryService, GroupV2Config, IdentId, IdentIdRef,
-    InboxOutcome, PayloadOutcome, RegistrationService,
+    ConversationId, ConvoMetadata, ConvoOutcome, Core, DeliveryService, GroupV2Config, IdentId,
+    IdentIdRef, InboxOutcome, PayloadOutcome, RegistrationService,
 };
 use logos_account::{AccountDirectory, resolve_device_ids};
 use parking_lot::Mutex;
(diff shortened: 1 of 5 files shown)


crates/generic-chat/src/client.rs
```diff
@@ -167,17 +167,22 @@ where
...
     /// group invite goes to every one of them. An empty slice creates a group
-    /// with only this client, to grow via [`Self::add_group_members`].
+    /// with only this client, to grow via [`Self::add_group_members`]. `name`
+    /// and `desc` become the group's shared metadata, carried to every joiner
+    /// in the welcome and readable via [`Self::group_metadata`]; both may be
+    /// empty.
     pub fn create_group_conversation(
         &mut self,
         accounts: &[AccountAddressRef],
+        name: &str,
```
kaichaosun: [suggestion] use a struct param for group metadata for future extensible.

kaichaosun (approved):

core/conversations/src/types.rs
```diff
@@ -72,3 +72,13 @@ pub struct ConvoMetadata {
...
 }
+
+impl ConvoMetadata {
+    /// Metadata for an unnamed group: both fields empty.
+    pub fn empty() -> Self {
+        Self {
+            name: String::new(),
+            desc: String::new(),
+        }
+    }
```
jazzz: [Dust] Empty should not be an encouraged state.

Its useful in the tests, however if its only used for testing consider making a generator function there.

All other use cases - developers having to be explicit about what values are being used would add clarity.


"All groups have a name" is a powerful invariant that is worth keeping. Even if developers choose ""

crates/generic-chat/src/client.rs
```diff
@@ -168,16 +168,20 @@ where
     /// account resolves to the signer ids its directory bundle endorses; the
     /// group invite goes to every one of them. An empty slice creates a group
     /// with only this client, to grow via [`Self::add_group_members`].
+    /// `metadata` becomes the group's shared name and description, carried to
+    /// every joiner in the welcome and readable via [`Self::group_metadata`];
+    /// both fields may be empty.
     pub fn create_group_conversation(
         &mut self,
         accounts: &[AccountAddressRef],
+        metadata: ConvoMetadata,
```
jazzz: [Sand] Use an discrete type, rather than reusing the output type ConvoMetaData.

Input metadata and output metadata will diverge as they are separate level concepts. ConvoMetaData could include conversation specific values such as Version and ConvoType - Its a method for ConvoTypes to expose relevant information to developers. 

Developers may not know/be able to generate some of these values. Consider decoupling  ConvoType -> Developer, Developer -> ConvoType pathways by using a separate type.

core/conversations/src/types.rs
jazzz: ref: https://github.com/logos-messaging/libchat/pull/161#discussion_r3541191824

jazzz (approved):

PR #188, opened 2026-07-23
feat: surface pending group invites on the roster

osmaczko:
A GroupV2 add is staged as a proposal that a later commit merges, so between the two the invited member sits in the conversation's own state and nowhere in its public roster, leaving a caller no way to observe an invite in flight.

group_members now returns those invites after the committed members, flagged pending. The flag is inviter-local and transient: a conversation records only the joiners this client proposed, and drops each one as the commit admitting it lands. GroupV1 never reports a pending member, since its add merges its own commit before returning.

### DIFF
--- a/core/conversations/src/core.rs
+++ b/core/conversations/src/core.rs
@@ -306,6 +306,24 @@ impl<'a, S: ExternalServices + 'static> Core<S> {
         }
     }
 
+    /// Each member invited here and still awaiting the group's commit, in the
+    /// same encoding as [`Self::group_members`]; errors if `convo_id` names a
+    /// direct (non-group) conversation.
+    pub fn group_pending_members(&mut self, convo_id: &str) -> Result<Vec<Vec<u8>>, ChatError> {
+        let convo = self
+            .cached_convos
+            .get(convo_id)
+            .ok_or_else(|| ChatError::NoConvo(convo_id.to_string()))?;
+
+        match convo {
+            ConvoTypeOwned::Group(group_convo) => group_convo.pending_members(),
+            ConvoTypeOwned::Direct(convo) => Err(ChatError::UnsupportedFunction(
+                convo.id().into(),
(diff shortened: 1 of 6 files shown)


core/conversations/src/core.rs
```diff
@@ -306,6 +306,29 @@ impl<'a, S: ExternalServices + 'static> Core<S> {
...
+
+            match convo {
+                ConvoTypeOwned::Group(group_convo) => group_convo.pending_members(),
+                ConvoTypeOwned::Direct(convo) => Err(ChatError::UnsupportedFunction(
+                    convo.id().into(),
+                    "List Pending Members".into(),
+                )),
+            }
+        } else {
+            let convo = self.load_group_convo(convo_id)?;
```
kaichaosun: [question] The `load_group_convo` looks almost not useful, it seems only used within groupv1. Not sure if we still want to maintain it as no one seems using groupv1.
Also `self.cached_convos.contains_key(convo_id)` is covered by `self.cached_convos.get(convo_id).ok_or_else`.

crates/generic-chat/tests/group_v2.rs
```diff
@@ -40,8 +40,18 @@ type TestClient = ChatClient<InProcessDelivery, EphemeralRegistry, ChatStorage>;
...
+) -> (TestClient, Receiver<Event>, String) {
+    create_test_client_with(message_bus, reg, fast_group_v2_config())
+}
+
+/// [`create_test_client`] with explicit GroupV2 timers, for a test that needs
+/// to observe the group between two protocol steps.
+fn create_test_client_with(
     message_bus: MessageBus,
     mut reg: EphemeralRegistry,
+    group_v2: GroupV2Config,
```
kaichaosun: [nitpick] better to use `config` as short name

kaichaosun (approved):

PR #189, opened 2026-07-23
feat: list a direct conversation's participants

osmaczko:
A direct conversation has participants like any conversation, and listing them is part of the conversation surface, but group_members could only answer for groups: the roster accessor sat on the group-only trait, so the direct arm returned an unsupported-function error even though the membership lives in the conversation's inner group.

members moves to the Convo trait and DirectV1Convo delegates it to that inner group. add_member stays on the group-only trait, so the direct arm keeps rejecting it.

### DIFF
--- a/core/conversations/src/core.rs
+++ b/core/conversations/src/core.rs
@@ -263,52 +263,36 @@ impl<'a, S: ExternalServices + 'static> Core<S> {
         convo_id: &str,
         members: &[IdentIdRef],
     ) -> Result<(), ChatError> {
-        if self.cached_convos.contains_key(convo_id) {
-            let convo = self
-                .cached_convos
-                .get_mut(convo_id)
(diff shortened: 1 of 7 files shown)


core/conversations/src/core.rs
```diff
@@ -545,4 +539,11 @@ impl<S: ExternalServices> Convo<S> for ConvoTypeOwned<S> {
             ConvoTypeOwned::Direct(convo) => convo.wakeup(service_ctx),
         }
     }
+
+    fn members(&self) -> Result<Vec<Vec<u8>>, ChatError> {
+        match self {
+            ConvoTypeOwned::Group(group_convo) => group_convo.members(),
```
kaichaosun, line 539: [nitpick] It looks like `self.members` is convenient way in this case, if not too much change of the abstraction.

kaichaosun (approved):

PR #191, opened 2026-07-27
feat(group_v2): event-driven liveness + found-with-members

seemenkina:
Two commits, kept separate:

**Drive liveness from de-mls events.** de-mls no longer keeps liveness timers — it emits events (`CommitWorkReady`, `UpdateRequestReceived`, `SyncResendNeeded`/`ConversationSyncObserved`, `ReelectionExhausted`) and the integrator decides when to act. This wires that on the libchat side: react to the event, arm a window, fire the trigger.

**Found a group with its members.** New `create_group_with_members` admits the founders in one genesis commit — one welcome, settled and stewarding from epoch 1, no consensus round — next to the existing proposal-per-member `create_group_conversation`.

Worth deciding on your side:

- `drive_liveness` collects every lever's anchor in one pass to keep the example simple. Split it per-lever, or drive off real presence/transport signals instead of blind timers, however fits you.
- Welcome delivery is inviter-only: if the inviter is offline when the commit lands, the joiner waits for a re-invite. A steward fallback would close that.
- `new_with_members` assumes an already-resolved, deduped participant list (no self). Filter first if you can't guarantee it, the way `add_member` dedups against live membership.

### DIFF
--- a/core/conversations/src/conversation/group_v2.rs
+++ b/core/conversations/src/conversation/group_v2.rs
@@ -114,6 +114,41 @@ fn rand_string(n: usize) -> String {
     hex::encode(bytes)
 }
 
+/// One fetched member: `(de-mls member_id, signer_id, key_package_bytes)`.
+type FetchedKeyPackage = (Vec<u8>, String, Vec<u8>);
+
+/// Fetch and dedupe each signer's key package, reading its de-mls member id from
+/// the KP leaf credential (de-mls matches by credential, not signer id). Errors
+/// if any member has no key package, before any are admitted.
+fn fetch_key_packages<S: ExternalServices>(
+    service_ctx: &ServiceContext<S>,
+    members: &[IdentIdRef],
+) -> Result<Vec<FetchedKeyPackage>, ChatError> {
+    let mut seen = std::collections::HashSet::new();
+    let mut invites = Vec::new();
+    for member in members
+        .iter()
+        .copied()
+        .filter(|m| seen.insert(m.as_str().to_string()))
+    {
+        let kp_bytes = service_ctx
+            .registry
+            .retrieve(member.as_str())
+            .map_err(ChatError::generic)?
+            .ok_or_else(|| ChatError::generic("No key package"))?;
+        let key_package_in = KeyPackageIn::tls_deserialize(&mut kp_bytes.as_slice())?;
+        let keypkg =
+            key_package_in.validate(service_ctx.mls_provider.crypto(), ProtocolVersion::Mls10)?;
+        let member_id = keypkg
+            .leaf_node()
+            .credential()
+            .serialized_content()
+            .to_vec();
+        invites.push((member_id, member.to_string(), kp_bytes));
+    }
+    Ok(invites)
+}
+
 fn group_config<S: ExternalServices>(
     cx: &mut ServiceContext<S>,
     name: &str,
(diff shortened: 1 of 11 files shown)


core/conversations/src/conversation/group_v2.rs
```diff
@@ -154,13 +195,72 @@ impl GroupV2Convo {
...
+    ///
+    /// Assumes `members` is already resolved and deduped, and excludes the
+    /// creator. An integrator that can't guarantee that should filter first, the
+    /// way `add_member` dedups against live membership.
+    pub fn new_with_members<S: ExternalServices>(
+        service_ctx: &mut ServiceContext<S>,
+        name: &str,
+        desc: &str,
+        members: &[IdentIdRef],
+    ) -> Result<Self, ChatError> {
```
jazzz, line 221: [!] This is good we'll look to deprecate `GroupV2::new` for this

core/conversations/src/conversation/group_v2.rs
```diff
@@ -154,13 +195,72 @@ impl GroupV2Convo {
...
+        let convo_id = rand_string(5);
+        let group_config = group_config(service_ctx, name, desc);
+        let invites = fetch_key_packages(service_ctx, members)?;
+        // `(member_id, key_package_bytes)` slices for the call.
+        let founders: Vec<(&[u8], &[u8])> = invites
+            .iter()
+            .map(|(member_id, _, kp)| (member_id.as_slice(), kp.as_slice()))
```
jazzz, line 228: [Dust] I personally find `founders` confusing; I'd assume thats the people who created the group rather than the 'initial_members`

core/conversations/src/conversation/group_v2.rs
```diff
@@ -417,33 +485,159 @@ where
...
+            }
+        }
+
+        // Recovery: mint each cycle while Layer-3 recovery is open.
+        if self.conversation.is_in_recovery_mode()
+            && let Err(e) = self.conversation.commit_in_recovery(provider, signer)
+        {
+            tracing::warn!(convo = %self.convo_id, error = %e, "commit_in_recovery failed");
+        }
+    }
```
jazzz, line 575: [Pebble]  This section is hard to follow; It appears to leak internal DeMLS state changes externally to ConversationTypes.

As a Consumer/maintainer I'm curious:
- How do I know where these values come from?
- How do I know when more if checks need to be performed?
- How can I prove this code is correct?

Incorrectly implementing this function looks like it would cause Conversations to silently fail. If LogosChat needs to be aware of the the internal state operations in-order to guarantee correct operation then the case for keeping DeMLS as a external library appears weakened.

core/conversations/src/conversation/group_v2.rs
```diff
@@ -417,33 +485,159 @@ where
...
+                }
+                ConversationEvent::ReelectionExhausted => {
+                    if let Err(e) = self
+                        .conversation
+                        .request_recovery(&service_ctx.mls_provider, &service_ctx.mls_identity)
+                    {
+                        tracing::warn!(convo = %self.convo_id, error = %e, "request_recovery failed");
+                    }
+                }
+                _ => {}
```
jazzz, line 634: [Sand] Hiding unhandled Events seems like a source of Program errors. 

Handling of types does not appear optional as it's required for nominal operation

jazzz (commented):
Member Initialization looks solid.

I have some questions about the liveness changes. Specifically around what is the intended API/Maintainence requirements.  These changes appear to change, the contract between the two libraries, and want to make sure I understand the direction we are heading.

core/conversations/src/conversation/group_v2.rs
seemenkina: I see. So my idea was to add more flexibility in term of recovery and reaction. But looks like i overdoing a little and move internal logic outside

> How do I know where these values come from?
> How do I know when more if checks need to be performed?
> How can I prove this code is correct?

So short answer you can't fully prove alone that code is correct. I can suggest to remove this from current PR and only left bug fixes and `new_with_members`. And do another research around how it can be done without leaking protocol decisions outside the de-mls

seemenkina:
Closed, will be reopen with only `new_with_members` feature. Experimental one going later in separate draft PR

PR #192, opened 2026-07-27
ci: switch the Nix cache to setup-nix-cache-action

igor-sirotin:
Same as logos-co/logos-chat-module#57 and logos-co/logos-delivery-module#69, using [`setup-nix-cache-action@v1`](https://github.com/logos-co/setup-nix-cache-action). Replaces `cache-nix-action` (per-runner GitHub cache) with the shared Attic cache; drops the `nix_version` pin since the action installs its own Nix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,21 +49,17 @@ jobs:
 
   smoketest:
     name: Smoketest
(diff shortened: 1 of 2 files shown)


jazzz (approved):
Looks like cuts the CI run time in half for the smoketests. Love it.

PR #194, opened 2026-07-29
feat(group_v2): create a group with its initial members in one commit

seemenkina:
When you create a group, everyone you picked now goes in together — one commit at creation, instead of adding them one by one after the group already exists. Backed by de-mls's `create_with_members`.

Also tidied the invite path along the way: deduped key-package fetch, best-effort add (a bad member is logged and skipped, not fatal), and `pending_invites` is a map keyed by member id.

upd: also fix time config in the tests, to small freeze window provide fork because we don't have enough time to get message trough the transport. I did't have such issue in de-mls because I use mock transport that not relate on real time so the timers from my repo doesn't satisfy real test application

### DIFF
--- a/core/conversations/src/conversation/group_v2.rs
+++ b/core/conversations/src/conversation/group_v2.rs
@@ -357,10 +357,6 @@ impl<S> GroupConvo<S> for GroupV2Convo
 where
     S: ExternalServices,
 {
-    // TODO(libchat): decide how to handle situation when add_member return error:
-    // 1) Break cycle and return
-    // 2) Skip this user, log or collect error and continue
-    // Also in additional to decision here, you also should check that fetch_key_packages reacts same way
     #[instrument(name = "groupv2.add_member", skip_all, fields(user_id = %service_ctx.mls_identity.display_name()))]
     fn add_member(
         &mut self,
@@ -372,6 +368,7 @@ where
         let members_to_add = fetch_key_packages(service_ctx, members)?;
         let existing: HashSet<Vec<u8>> = self.conversation.members()?.into_iter().collect();
 
+        let mut result = Ok(());
         for FetchedMember {
             member_id,
             signer,
@@ -381,23 +378,23 @@ where
             if existing.contains(&member_id) {
                 continue;
             }
-            match self.conversation.add_member(
+            self.pending_invites.insert(member_id.clone(), signer);
+            if let Err(e) = self.conversation.add_member(
                 &service_ctx.mls_provider,
                 &service_ctx.mls_identity,
                 &member_id,
                 &key_package,
             ) {
-                // Record the invite only on success
-                Ok(()) => {
-                    self.pending_invites.insert(member_id, signer);
-                }
-                // One bad member doesn't block the rest.
-                Err(e) => tracing::warn!(member = %signer, error = %e, "could not add member"),
+                self.pending_invites.remove(&member_id);
+                result = Err(e.into());
+                break;
             }
         }
-
-        self.after_op(service_ctx)?;
-        Ok(())
+        // Flush even on a mid-loop failure: proposals already opened must be
+        // published and the wakeup re-armed, or they sit dormant until an
+        // unrelated frame drives the conversation.
+        let flushed = self.after_op(service_ctx).map(drop);
+        result.and(flushed)
     }
 
     fn pending_members(&self) -> Result<Vec<Vec<u8>>, ChatError> {
--- a/core/conversations/src/conversation/group_v2.rs
+++ b/core/conversations/src/conversation/group_v2.rs
@@ -25,6 +25,7 @@ use openmls::prelude::tls_codec::Deserialize as _;
 use openmls::prelude::{KeyPackageIn, OpenMlsProvider as _, ProtocolVersion};
 use prost::Message;
 use shared_traits::{IdentId, IdentIdRef};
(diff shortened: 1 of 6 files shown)


core/conversations/src/conversation/group_v2.rs
```diff
@@ -302,91 +357,51 @@ impl<S> GroupConvo<S> for GroupV2Convo
...
+                &key_package,
             ) {
-                self.pending_invites.pop();
-                result = Err(e.into());
-                break;
+                // Record the invite only on success
+                Ok(()) => {
+                    self.pending_invites.insert(member_id, signer);
+                }
+                // One bad member doesn't block the rest.
```
osmaczko: I'd go with option 1, break and return the error.

I don't think the skip-and-continue case can happen today anyway: a member without a key package is already rejected by `fetch_key_packages` before any proposal opens, so what's left here is conversation-wide (`ConversationBlocked`, `PartialFreeze`) and the same for every member. So the warn arm doesn't really skip a bad member, it just turns a total failure into `Ok(())` where the caller can't tell 3-of-3 from 0-of-3.

And if per-member failures do show up later, I think break-and-return still reports honestly, while the warn arm would quietly start producing partial adds.

core/conversations/src/conversation/group_v2.rs
```diff
@@ -88,10 +89,10 @@ fn make_consensus() -> DefaultConsensusPlugin {
 pub struct GroupV2Convo {
     convo_id: String,
     conversation: Conversation<DefaultConsensusPlugin, InMemoryPeerScoreStorage, GroupV2Clock>,
-    /// Joiners WE invited, as `(member_id, signer_id)`: the de-mls member id
-    /// (the joiner's leaf credential content, read from its key package) paired
-    /// with the signer id its welcome is delivered to.
-    pending_invites: Vec<(Vec<u8>, String)>,
+    /// Joiners WE invited, keyed by de-mls member id (the joiner's leaf
+    /// credential content, read from its key package) → the signer id its
+    /// welcome is delivered to.
```
jazzz, line 94: [!] The credential must be treated as opaque bytes and does not necessarily uniquely identify the member. While the current implementation this works, it will break in  #195.

core/conversations/src/conversation/group_v2.rs
```diff
@@ -124,14 +125,62 @@ fn group_config(name: &str, desc: &str) -> MlsGroupCreateConfig {
...
+/// credential content, which de-mls matches on), the `signer` its welcome
+/// routes to, and the `key_package` bytes to commit.
+struct FetchedMember {
+    member_id: Vec<u8>,
+    signer: IdentId,
+    key_package: Vec<u8>,
+}
+
+/// Fetch and dedupe each signer's key package, reading its de-mls member id from
+/// the KP leaf credential (de-mls matches by credential, not signer id). Errors
```
jazzz, line 138: > de-mls matches by credential, not signer id

[?] Any details on this? I see potential conflicts with the Account restructuring work. Is this a necessity? or just how it functions at this time?

My Concern is that the Credential is user supplied, and cannot be validated without access to the AuthenticationService( currently at the Client-level).

core/conversations/src/conversation/group_v2.rs
jazzz: ref:

~line: 360

> // TODO(libchat): decide how to handle situation when add_member return error:
    // 1) Break cycle and return
    // 2) Skip this user, log or collect error and continue
    // Also in additional to decision here, you also should check that fetch_key_packages reacts same way

jazzz (commented):
No strict blockers,

But need to clear up questions on the canonical member_id definition between DeMLS and LibChat.

core/conversations/src/conversation/group_v2.rs
seemenkina: cool, I'll change the code to satisfy it and remove TODO

core/conversations/src/conversation/group_v2.rs
seemenkina: Necessity, not just how it works now — full reasoning in the [de-mls thread](https://github.com/vacp2p/de-mls/pull/143#discussion_r3689964243). Short version:

de-mls keys its durable cross-epoch state (steward lists, scores, dedup) on member_id = credential.serialized_content(). Per RFC 9420 §5.3.3 the credential is the only identifier that's both persistent and authenticated — signature_key / leaf index are unique but only per-epoch and they rotate, so they can't anchor state that must survive rotation.

On the AS concern — yes, intentional. de-mls has no AS and treats the credential as opaque bytes. It only verifies possession of the leaf's signing key + current membership, not that the credential really belongs to the claimed party — that stays at the Client-level AS in libchat. So de-mls basically believes the credential was already AS-verified before the add. Ofc it can catch some cases on top (e.g. a credential collision inside the group), but identity↔key auth is more about the AS.

On #195 — the one hard thing de-mls needs: credential.serialized_content() must be byte-stable and the same per member (we compare ids by raw bytes, no normalization), or steward/score continuity breaks. If one account maps to several leaves, I think we need a per-device credential so each leaf keeps a unique + stable member_id. Let's align on the encoding.

core/conversations/src/conversation/group_v2.rs
seemenkina: addressed this concern below

jazzz (approved):
Nothing blocking in here; looks good!

PR #197, opened 2026-08-04
fix(build): re-run logos-delivery build script when the native lib apears

mchenani:
Fixes `chat-cli` failing to link with undefined `_logosdelivery_*` symbols when the native lib is built after the first `cargo build`. The build script now watches the `LOGOS_DELIVERY_LIB_DIR` path so it re-runs once the lib appears, making build order irrelevant; the fallback `nix` build also enables flakes inline, and the `README` leads with nix develop.

### DIFF
--- a/bin/chat-cli/README.md
+++ b/bin/chat-cli/README.md
@@ -4,8 +4,15 @@ A terminal chat application built on top of libchat. End-to-end encrypted messag
 
 ## Building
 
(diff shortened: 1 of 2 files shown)


jazzz (approved):
🚀  - Great fix. 

This area has become out of date after #176.  But as the first place new contributors look, it needs to be a clean example. Thanks for the help!

